PR #80, opened 2026-04-29
[nojira][fix] Tighten PR workflow guards

VytCepas:
No linked issue (nojira).


VytCepas:
**Review response:**
- Guard blocks issue-backed PR creation when start-issue.sh is absent: fixed. The guard now blocks raw issue-backed PR creation only when start-issue.sh exists, and blocks raw nojira PR creation only when the command is actually creating a [nojira] PR.
- monitor-pr.sh waits for review before reporting CI failures: fixed. It now parses current gh pr checks fields (state/bucket), reports CI failures immediately after checks settle, and only waits for review when CI is green.
- Stacked PR diff: fixed. Rebasing onto origin/main removed the earlier PI-30 commit from this PR.

PR #81, opened 2026-04-30
[nojira][fix] Reduce review cycle wait

VytCepas:
No linked issue (nojira).


VytCepas:
**Review response:**
- Header still said review-cycle N >= 2: fixed in both the active and template monitor-pr.sh headers by tying the wording to MAX_REVIEW_CYCLES instead of a stale hard-coded value.

PR #185, opened 2026-06-14
fix(PI-183): upload only build artifacts; harden the pypi environment

### LINKED ISSUE #183: Release workflow attaches dist/.gitignore as a stray default.gitignore asset
## Summary

The `Release` workflow attaches an extra, unwanted asset named `default.gitignore` to every GitHub Release, alongside the wheel and sdist. Seen on the v0.3.0 release:

> assets: **default.gitignore**, project_init-0.3.0-py3-none-any.whl, project_init-0.3.0.tar.gz

## Cause

`release.yml` uploads `files: dist/*`. The repo tracks a `dist/.gitignore` (so the `dist/` dir exists in git), which is present in the checkout when `uv build` runs into `dist/`. The glob therefore picks it up, and `softprops/action-gh-release` surfaces the dotfile as `default.gitignore`.

```yaml
# .github/workflows/release.yml — Create GitHub Release step
with:
  files: dist/*          # <-- also matches dist/.gitignore
```

## Fix

Narrow the glob to the build artifacts only:

```yaml
with:
  files: |
    dist/*.whl
    dist/*.tar.gz
```

Applies to both the `release` job (GitHub Release assets). The `publish-pypi` job is unaffected (`uv build` + `pypa/gh-action-pypi-publish` ignore the dotfile).

## Severity

Minor / cosmetic — a junk asset on the release page; no functional impact on installs or PyPI.

Found while publishing v0.3.0.

### LINKED ISSUE #184: Harden the pypi deployment environment with protection rules
## Summary

Reminder to properly configure the `pypi` GitHub **deployment environment** that the release pipeline publishes through.

## Context

During the v0.3.0 release the `publish-pypi` job (`release.yml`) targets `environment: pypi`, but that environment did not exist in the repo yet (only `copilot` did). It was created ad-hoc via the API to unblock the publish:

```
gh api -X PUT repos/VytCepas/project-init/environments/pypi
```

That created a **bare, unprotected** environment — no deployment branch/tag policy and no required reviewer. The first publish worked (trusted-publishing security comes from the OIDC match on PyPI's side: owner/repo/workflow/environment), but the GitHub environment itself should be hardened.

## What to do later

In **Settings → Environments → `pypi`**, add defense-in-depth so the publish path can't be triggered off arbitrary refs:

- [ ] **Deployment branches and tags** → restrict to tags matching `v*` (the release only ever runs on `v*` tags).
- [ ] (Optional) **Required reviewers** → yourself, so a publish pauses for a manual click before uploading to PyPI.
- [ ] (Optional) Document the environment + trusted-publisher setup in the release runbook / ADR-011 so it's reproducible and not just tribal knowledge.

## Severity

Low — the environment exists and publishing works; this is hardening, not a blocker.

Follow-up from the v0.3.0 release (#171).

VytCepas:
Two release-infra follow-ups from the v0.3.0 publish.

## PI-183 — stray release asset
The `release` job used `files: dist/*`, which also globbed the tracked `dist/.gitignore` and attached it as a `default.gitignore` asset (visible on the v0.3.0 release). Narrowed to `dist/*.whl` + `dist/*.tar.gz`. Contract test updated to assert the artifacts and forbid the bare glob.

## PI-184 — harden the `pypi` deployment environment
The `pypi` environment was created bare to unblock the v0.3.0 publish. It is now restricted to `v*` tags (applied via the API), so only a tagged release can deploy to it — defense in depth on top of the PyPI-side OIDC trusted-publisher identity match. ADR-011 updated to record the protection and the completed one-time publisher setup. A required-reviewer approval gate can be added later if wanted (noted in the ADR).

550 tests pass, ruff clean, `release.yml` validated as YAML.

Closes #183
Closes #184

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,9 +50,9 @@ jobs:
       - name: Create GitHub Release
         uses: softprops/action-gh-release@v3
         with:
-          # Only the build artifacts — a bare `dist/*` also globs the tracked
-          # dist/.gitignore and uploads it as a stray `default.gitignore` asset
-          # (PI-183).
+          # Only the build artifacts — an unscoped glob also picks up the
+          # tracked dist/.gitignore and uploads it as a stray
+          # `default.gitignore` asset (PI-183).
           files: |
             dist/*.whl
             dist/*.tar.gz
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,7 +50,12 @@ jobs:
       - name: Create GitHub Release
         uses: softprops/action-gh-release@v3
         with:
-          files: dist/*
+          # Only the build artifacts — a bare `dist/*` also globs the tracked
+          # dist/.gitignore and uploads it as a stray `default.gitignore` asset
+          # (PI-183).
+          files: |
+            dist/*.whl
+            dist/*.tar.gz
           body_path: RELEASE_NOTES.md
 
   # PyPI via trusted publishing (ADR-011): OIDC, no tokens to custody.
--- a/docs/adr/adr-011-pypi-trusted-publishing.md
+++ b/docs/adr/adr-011-pypi-trusted-publishing.md
@@ -28,7 +28,9 @@ Publish every tagged release to PyPI from the existing `release.yml` via a
 
 - Runs only after the GitHub Release job succeeds (`needs: release`).
 - Authenticates via OIDC (`id-token: write`) against the `pypi`
-  environment — protection rules can be added on the environment later.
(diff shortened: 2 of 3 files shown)


VytCepas:
**Review response:** Good catch — fixed in the latest commit. Replaced the narrow `"files: dist/*\n" not in content` check with a regex (`dist/\*(?!\.(?:whl|tar\.gz))`) that flags any bare `dist/*` regardless of trailing whitespace, comment, or EOF. Also reworded the release.yml comment so it no longer contains the literal pattern (which the stricter regex would otherwise flag). 550 tests pass.

PR #211, opened 2026-06-17
fix(PI-207): repair board automation on user-owned repos (resolver + repo pin)

### LINKED ISSUE #207: board-automation.yml fails on personal (user-owned) accounts: org+user GraphQL query trips set -e
## Bug: `board-automation.yml` fails on personal (user-owned) accounts

**Scaffold version:** project-init 0.1.0, `obsidian-only` preset, `language: python`.

### Symptom
On a repo owned by a **personal account** (not an organization), the `Board automation` workflow's `Sync project item fields` step fails:

```
gh: Could not resolve to an Organization with the login of '<user>'.
##[error]Process completed with exit code 1.
```

This happens on every issue/PR event once `PROJECT_TOKEN` is configured, so the board never updates. (Note: this is only reached after `PROJECT_TOKEN` is a classic PAT with `project` scope — fine-grained PATs separately fail to *write* to user-owned Projects v2 with "Resource not accessible by personal access token".)

### Root cause
The project-lookup query in `.github/workflows/board-automation.yml` requests **both** the `user(login:$owner)` and `organization(login:$owner)` project paths in a single GraphQL call:

```graphql
query($owner: String!, $number: Int!) {
  user(login: $owner)         { projectV2(number: $number) { ... } }
  organization(login: $owner) { projectV2(number: $number) { ... } }
}
```

For a personal account, the `organization` path returns a GraphQL error ("Could not resolve to an Organization"). `gh api graphql` exits **non-zero** whenever the response contains an `errors` array — even though the response still carries valid `data.user.projectV2`. Because the step runs under `set -euo pipefail`, the non-zero exit kills the step at the `PROJECT_DATA=$(...)` assignment, before the existing `jq '.data.user.projectV2 // .data.organization.projectV2 // empty'` fallback (line ~129) ever runs. (Symmetrically, an org-owned repo would error on the `user` path.)

### Suggested fix
Tolerate the partial-error response so the `jq` fallback can select whichever path applies:

```bash
PROJECT_DATA=$(gh api graphql -f query='...' -f owner="$OWNER" -F number="$PROJECT_NUMBER" 2>/dev/null || true)
```

The downstream `jq // empty` + `PROJECT_ID` empty-check already handle the "neither path matched" case gracefully. Alternatively, split into two separate queries, or detect the owner type first.

---
**Related defects found on the same fresh scaffold (v0.1.0):** #208 (CI Python version / matrix), #209 (`uv sync --extra dev` vs `[dependency-groups]`), #210 (`validate-pr.yml` missing `GH_REPO`).

---

## Planning metadata
- Type: bug
- Priority: high
- Size: S
- Area: board-automation.yml (template)
- Agent ready: Yes
- Confidence: high

## References
- #208, #209, #210

## Dependencies
- None

## Acceptance criteria
- Board automation works on personal user-owned accounts, not only organizations.

## Definition of Ready

## Definition of Done

### LINKED ISSUE #234: board-automation.yml never succeeds on user-owned repos (org/user resolver aborts under set -e; gh issue view has no checkout)
## Summary

`board-automation.yml` (the scaffolded GitHub Projects sync workflow) **never succeeds on a user-owned repository**. Two independent bugs each abort the `board-sync` job before it can add an issue to the board or sync any field. On an org-owned repo with the default checkout assumptions it happens to work, which is why this has gone unnoticed.

Discovered while scaffolding `VytCepas/zarija` (personal account). Every `issues`/`pull_request` event produced a failed `Board automation` run and an empty project board; items had to be added manually. Both bugs are in the template, not the consuming repo.

## Environment

- Repo owner type: **personal user account** (not an organization)
- `PROJECT_TOKEN` correctly configured with `project` scope (not the cause)
- Project v2 number set (`PROJECT_NUMBER: 5`)

---

## Bug 1 — project resolver aborts under `set -euo pipefail` on personal accounts

The resolver probes both `user(login:)` and `organization(login:)` in a single GraphQL query so it can support either owner type, then picks the non-null one via jq:

```bash
PROJECT=$(echo "$PROJECT_DATA" | jq -c '.data.user.projectV2 // .data.organization.projectV2 // empty')
```

On a personal account, the `organization(login: "<user>")` half resolves to **nothing** and GitHub returns **HTTP 200 with partial `data` PLUS a top-level `errors` array** (`Could not resolve to an Organization with the login of '<user>'`). `gh api graphql` treats any `errors` array as failure and **exits non-zero**, so under `set -euo pipefail` the command-substitution assignment aborts the entire step — before the `addProjectV2ItemById` mutation ever runs. (Symmetrically, an org account would trip the `user(login:)` half.)

### Fix 1

Swallow the expected partial-error exit so the existing jq fallback can select whichever owner type actually resolved:

```diff
-            }' -f owner="$OWNER" -F number="$PROJECT_NUMBER")
+            }' -f owner="$OWNER" -F number="$PROJECT_NUMBER" 2>/dev/null || true)
```

The needed `data` is still on stdout; only the error needs suppressing. The downstream `if [ -z "$PROJECT_ID" ]; then ... "project not found or PROJECT_TOKEN lacks access" ... fi` guard still handles a genuine hard failure gracefully.

---

## Bug 2 — `gh issue view` fails with no checkout (masked by Bug 1)

The `board-sync` job has **no `actions/checkout` step** (by design — it only needs API access). But it calls:

```bash
ISSUE_DATA=$(gh issue view "$ISSUE_NUM" --json labels,body)
```

`gh issue view` resolves the target repo from the **local git remote**, which doesn't exist without a checkout, so it dies with:

```
failed to run git: fatal: not a git repository (or any of the parent directories): .git
```

This was latent — Bug 1 aborted the job earlier, so it only surfaced once Bug 1 was fixed.

### Fix 2

Pin the repo for every `gh` call via the job-level `env` (cleaner than adding `-R` to each command, and avoids needing a checkout):

```diff
     env:
       GH_TOKEN: ${{ secrets.PROJECT_TOKEN }}
+      # No actions/checkout in this job; pin the repo so gh commands like
+      # `gh issue view` work checkout-free instead of failing on a missing
+      # git remote.
+      GH_REPO: ${{ github.repository }}
       OWNER: ${{ github.repository_owner }}
       REPO_NAME: ${{ github.event.repository.name }}
       PROJECT_NUMBER: 5
```

---

## Verification (done downstream in VytCepas/zarija)

With both fixes merged: creating an issue fires `board-sync` → **success**, the issue auto-adds to the board as **To Do**, and merging a PR with `Closes #N` flips the linked issue to **Done**. Confirmed green on a personal-account repo.

## Suggested template change

Apply both diffs to the `board-automation.yml` template. Optionally add a one-line note in the workflow header that the job is intentionally checkout-free and supports both user- and org-owned repos.

VytCepas:
## Summary

Repairs `board-automation.yml` so the project-board sync works on **user-owned** repos. Two bugs (both reported in #234; Bug 1 is the original #207):

### Bug 1 — project resolver aborts under `set -euo pipefail` (PI-207)
The lookup queries both `user(login:)` and `organization(login:)` in one GraphQL call. On a personal account the `organization` half returns HTTP 200 with `data` **plus** an `errors` array, so `gh api graphql` exits non-zero and the command-substitution assignment aborts the step before any board mutation runs. **Fix:** tolerate the partial-error (`2>/dev/null || true`) so the existing `jq '.data.user.projectV2 // .data.organization.projectV2 // empty'` fallback selects whichever owner type resolved.

### Bug 2 — gh needs repo context with no checkout (PI-234)
The board-sync job has no `actions/checkout` (by design). The `gh issue view` call has been pinned via `-R "$OWNER/$REPO_NAME"` since PI-111, so it already works — but per the reporter's suggestion this adds the cleaner **env-level** pin so *every* gh call in the job resolves the repo checkout-free:
```yaml
GH_REPO: ${{ github.repository }}
```

## Test
`test_board_automation_tolerates_personal_accounts` asserts both the partial-error tolerance and the repo pin.

Closes #207
Closes #234

### DIFF
--- a/templates/base/dot_github/workflows/board-automation.yml
+++ b/templates/base/dot_github/workflows/board-automation.yml
@@ -118,9 +118,15 @@ jobs:
                   }
                 }
               }
-            }' -f owner="$OWNER" -F number="$PROJECT_NUMBER")
+            }' -f owner="$OWNER" -F number="$PROJECT_NUMBER" 2>/dev/null || true)
 
-          PROJECT=$(echo "$PROJECT_DATA" | jq -c '.data.user.projectV2 // .data.organization.projectV2 // empty')
+          # A personal (user-owned) account errors on the organization(...) path
+          # (and an org errors on the user(...) path); gh exits non-zero whenever
+          # the response carries an `errors` array, even though valid data.user /
+          # data.organization is still on stdout. Tolerate that so the jq fallback
+          # below can select whichever path applies. Total failures leave
+          # PROJECT_DATA empty -> PROJECT_ID empty -> graceful skip. (PI-207)
+          PROJECT=$(echo "$PROJECT_DATA" | jq -c '.data.user.projectV2 // .data.organization.projectV2 // empty' 2>/dev/null || true)
           PROJECT_ID=$(echo "$PROJECT" | jq -r '.id // empty')
 
           if [ -z "$PROJECT_ID" ]; then
--- a/templates/base/dot_github/workflows/board-automation.yml
+++ b/templates/base/dot_github/workflows/board-automation.yml
@@ -24,6 +24,10 @@ jobs:
     runs-on: ubuntu-24.04
     env:
       GH_TOKEN: ${{ secrets.PROJECT_TOKEN }}
+      # No actions/checkout in this job (it only needs API access); pin the repo
+      # so every `gh` call (e.g. `gh issue view`) works checkout-free instead of
+      # failing on a missing git remote (PI-234).
+      GH_REPO: ${{ github.repository }}
       OWNER: ${{ github.repository_owner }}
       REPO_NAME: ${{ github.event.repository.name }}
       PROJECT_NUMBER: 1
--- a/tests/contracts/test_templates.py
+++ b/tests/contracts/test_templates.py
@@ -128,6 +128,12 @@ class TestScaffoldGitHubFiles:
     def test_board_automation_workflow_created(self):
         assert (self.target / ".github" / "workflows" / "board-automation.yml").is_file()
 
+    def test_board_automation_tolerates_personal_accounts(self):
+        """PI-207: the user+organization project query errors on one path; the
+        lookup must tolerate that so the jq fallback selects whichever applies."""
+        content = (self.target / ".github" / "workflows" / "board-automation.yml").read_text()
+        assert "2>/dev/null || true)" in content
+
     def test_review_status_workflow_created(self):
         assert (self.target / ".github" / "workflows" / "review-status.yml").is_file()
 
--- a/tests/contracts/test_templates.py
+++ b/tests/contracts/test_templates.py
@@ -129,10 +129,12 @@ class TestScaffoldGitHubFiles:
         assert (self.target / ".github" / "workflows" / "board-automation.yml").is_file()
 
     def test_board_automation_tolerates_personal_accounts(self):
-        """PI-207: the user+organization project query errors on one path; the
-        lookup must tolerate that so the jq fallback selects whichever applies."""
+        """PI-207/PI-234: the user+organization project query errors on one path
+        (tolerate it so the jq fallback selects whichever applies), and the
+        checkout-free job must pin the repo so gh calls work without a remote."""
         content = (self.target / ".github" / "workflows" / "board-automation.yml").read_text()
         assert "2>/dev/null || true)" in content
+        assert "GH_REPO:" in content
 
     def test_review_status_workflow_created(self):
         assert (self.target / ".github" / "workflows" / "review-status.yml").is_file()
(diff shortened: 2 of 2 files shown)


VytCepas:
**Review response (Copilot):** Fixed in `aefa3f9`. The `test_board_automation_tolerates_personal_accounts` contract test no longer asserts the bare `"2>/dev/null || true)"` substring against the whole workflow file (brittle to whitespace/formatting, and could match the unrelated `PROJECT=$(... || true)` line). It now locates the specific multi-line `PROJECT_DATA=$(gh api graphql ...)` command substitution — capturing it through the line that closes the substitution — and asserts that block carries the gh project query (`gh api graphql`, `projectV2`) plus the error-suppression tokens (`2>/dev/null` and `|| true`) checked as separate tokens so spacing changes can't break it. Original intent is preserved: board automation must tolerate personal accounts where the org Projects API errors (PI-207); the `GH_REPO:` pin check (PI-234) is retained. `uv run pytest tests/contracts/test_templates.py` (37 passed) and `uv run ruff check` both pass.

PR #213, opened 2026-06-17
fix(PI-208): stop pinning CI to Python 3.12; let uv resolve the version

### LINKED ISSUE #208: ci.yml hardcodes Python 3.12, conflicts with requires-python >=3.13 — support a version matrix instead
## Bug + enhancement: `ci.yml` hardcodes a single Python version that conflicts with `requires-python`

**Scaffold version:** project-init 0.1.0, `language: python`.

### Symptom
On a fresh scaffold, the CI `Lint and test` job fails at the dependency-sync step:

```
error: The requested interpreter resolved to Python 3.12.x, which is incompatible
with the project's Python requirement: `>=3.13` (from `project.requires-python`).
##[error]Process completed with exit code 2.
```

### Root cause
`ci.yml` pins the interpreter to one version in two places:

- `astral-sh/setup-uv` step: `python-version: "3.12"`
- integration-tests job: `uv python install 3.12`

…while the generated `pyproject.toml` sets `requires-python = ">=3.13"` (matching the generated `.python-version`). The pinned 3.12 is below the floor, so `uv sync` refuses to resolve.

### Recommendation: drive CI from a version matrix, don't lock to one interpreter
Rather than hardcoding a single version (and then having to keep it in sync with `requires-python`), the scaffold should test across **multiple** supported Python versions via a matrix. This avoids locking the project to one interpreter and surfaces version-specific dependency-resolution problems early, instead of implicitly tying the project to whatever single version CI happens to pin.

Suggested shape:

```yaml
jobs:
  lint-and-test:
    strategy:
      fail-fast: false
      matrix:
        python-version: ["3.13", "3.14"]   # derived from / consistent with requires-python
    steps:
      - uses: astral-sh/setup-uv@v8.1.0
        with:
          python-version: ${{ matrix.python-version }}
```

The matrix floor must respect `requires-python`. Ideally the scaffold derives the matrix from the chosen `requires-python` at init time so the two can never drift. As a minimum bug fix, the hardcoded `3.12` should at least match the `requires-python` floor.

Related: this was found alongside #207 (board automation) and the dev-dependency-sync and validate-pr issues filed separately.

---

## Planning metadata
- Type: bug
- Priority: high
- Size: M
- Area: ci.yml.tmpl (template)
- Agent ready: Yes
- Confidence: high

## References
- #209

## Dependencies
- None

## Acceptance criteria
- CI runs a Python version matrix that respects requires-python and cannot drift from it.

## Definition of Ready

## Definition of Done

VytCepas:
## Summary

`ci.yml` hardcoded the interpreter in two places — `astral-sh/setup-uv` `python-version: "3.12"` and `uv python install 3.12` — which is **below** a scaffold that sets `requires-python >= 3.13`. `uv sync` then refused to resolve and CI failed on a fresh scaffold.

## Fix

Drop both pins. `uv sync` installs the version the project requires (from `.python-version` / `requires-python`), so CI tracks the project and can never pin below the floor.

> The **multi-version matrix** enhancement requested in the issue is a follow-up (filed separately) — a real derived matrix needs a scaffold variable so the version set can't drift from `requires-python`. This PR fixes the breakage; the comment in the template marks where a matrix would slot in.

## Test

Adds `test_ci_does_not_hardcode_python_version` (scaffold-into-temp contract test).

Closes #208

### DIFF
--- a/templates/base/dot_github/workflows/ci.yml.tmpl
+++ b/templates/base/dot_github/workflows/ci.yml.tmpl
@@ -26,7 +26,9 @@ jobs:
           version: "0.11.7"
           enable-cache: true
           cache-dependency-glob: "pyproject.toml"
-          python-version: "3.12"
+          # No python-version pin (PI-208): `uv sync` installs the version the
+          # project requires (.python-version / requires-python), so CI never
+          # drifts below the floor. Add a matrix here to test multiple versions.
 
       # CI invokes the same justfile recipes agents and humans use (PI-139),
       # so the command surface is defined exactly once.
@@ -141,9 +143,8 @@ jobs:
           enable-cache: true
           cache-dependency-glob: "pyproject.toml"
 
-      - name: Set up Python
-        run: uv python install 3.12
-
+      # uv sync installs the project's required Python (.python-version /
+      # requires-python) — no version to drift (PI-208).
       - name: Sync dev dependencies
         run: uv sync --extra dev
 
--- a/tests/contracts/test_templates.py
+++ b/tests/contracts/test_templates.py
@@ -151,10 +151,24 @@ class TestScaffoldGitHubFiles:
 
     def test_ci_does_not_hardcode_python_version(self):
         """PI-208: a pinned Python version drifts below requires-python; let uv
-        resolve the project's interpreter from .python-version/requires-python."""
+        resolve the project's interpreter from .python-version/requires-python.
+
+        Version-agnostic (Copilot review): guards against *any* hardcoded pin,
+        not just 3.12, so reintroducing 3.13 or `uv python install <x>` also
+        fails. A version matrix (list or ${{ matrix.* }} reference) stays
+        allowed — neither matches a bare scalar literal.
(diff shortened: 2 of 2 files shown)


VytCepas:
**Review response (Copilot):** Fixed in 5029d2a — the check is now version-agnostic. The previous assertions only caught a literal `3.12` and would still have passed if CI pinned another version (e.g. `3.13`) or reintroduced `uv python install` with a different version. `test_ci_does_not_hardcode_python_version` now rejects *any* hardcoded pin via regex:

- `uv python install \d` — no `uv python install <version>` for any version
- `python-version:\s*["']?\d` — no literal `python-version:` scalar pin (a version matrix list or `${{ matrix.* }}` reference stays allowed)

This keeps the PI-208 intent intact: CI must not hard-pin Python; it relies on `requires-python` / `.python-version` resolved by `uv sync`. Verified locally: `uv run pytest tests/contracts/test_templates.py` (37 passed) and `ruff check` both green.

PR #216, opened 2026-06-17
fix(PI-186): bound the monitor_pr CI-wait loop so it can't hang forever

### LINKED ISSUE #186: monitor_pr.sh CI-wait loop is unbounded — autonomous merge can hang indefinitely
### What happened
`monitor_pr.sh` bounds the **review** wait (`REVIEW_TIMEOUT=360`, applied at line 188) but the **CI** wait loop has no cap:

```sh
# templates/base/dot_claude/scripts/monitor_pr.sh:150
while true; do
  ...
  [ "$CHECK_COUNT" -eq 0 ] && { sleep 10; continue; }
  PENDING=$(_count_pending "$CHECKS")
  [ "$PENDING" -eq 0 ] && break
  sleep 10
done
```

If a required check never leaves `PENDING`/`EXPECTED`/`IN_PROGRESS` (e.g. a workflow that never triggers on this branch, or a check required by branch protection but never registered), the loop spins forever. In a fully autonomous `--merge` run this hangs the agent driving it, with no timeout and no diagnostic. It is the one place the script's otherwise-pervasive fail-safe discipline lapses.

### Steps to reproduce
1. Open a PR whose branch protection requires a check that does not actually run on the branch.
2. Run `.claude/scripts/monitor_pr.sh <pr> --merge`.
3. The script blocks indefinitely in the CI-wait loop.

### Expected behaviour
The CI wait should have a timeout / iteration cap (mirroring `REVIEW_TIMEOUT`), then fail-closed with a clear "CI did not settle in Ns" message, a non-zero exit, and the names of the still-pending checks.

### Area / component
templates (scaffolded `monitor_pr.sh`); mirror the fix in the repo copy `.claude/scripts/monitor_pr.sh`.

_Found in internal code review (project-init self-audit). Agent ready: Yes._

---

## Planning metadata
- Type: bug
- Priority: high
- Size: S
- Area: monitor_pr.sh (template + root copy)
- Agent ready: Yes
- Confidence: high

## References
- #203

## Dependencies
- None

## Acceptance criteria
- CI-wait loop is bounded by a timeout and fails closed, naming the stuck checks.

## Definition of Ready

## Definition of Done

VytCepas:
## Summary

`monitor_pr.sh` caps the **review**-wait loop at `REVIEW_TIMEOUT` (360s), but the **CI**-wait loop was `while true` with no cap. A required check that never leaves `PENDING`/`EXPECTED`/`IN_PROGRESS` — a workflow that never triggers on the branch, or a branch-protection check that never registers — made the script (and any autonomous agent driving `--merge`) spin forever with no diagnostic.

## Fix

Add `CI_TIMEOUT` (900s) with elapsed tracking, mirroring the review loop. On timeout, print the still-pending check names and `exit 1` (fail closed) rather than hang. The "checks not registered yet" guard is preserved (now bounded).

## Test

Adds `test_monitor_pr_ci_wait_is_bounded` (scaffold-into-temp contract test).

Closes #186

### DIFF
--- a/templates/base/dot_claude/scripts/monitor_pr.sh
+++ b/templates/base/dot_claude/scripts/monitor_pr.sh
@@ -147,13 +147,28 @@ _has_review_activity() {
 # Guard: if checks haven't registered yet (empty list), keep polling.
 # An empty list is indistinguishable from "all done" without this guard,
 # which caused premature merges before CI even started.
+# Bounded wait (PI-186): a required check that never leaves PENDING/EXPECTED
+# (e.g. a workflow that never triggers on this branch) must not hang the
+# script — and any autonomous caller — forever. Fail closed on timeout.
+CI_TIMEOUT=900
+CI_ELAPSED=0
 while true; do
   CHECKS=$(gh pr checks "$PR_NUMBER" --json name,state,bucket 2>/dev/null) || CHECKS="[]"
   CHECK_COUNT=$(echo "$CHECKS" | python3 -c "import json,sys; print(len(json.load(sys.stdin)))")
-  [ "$CHECK_COUNT" -eq 0 ] && { sleep 10; continue; }
-  PENDING=$(_count_pending "$CHECKS")
-  [ "$PENDING" -eq 0 ] && break
+  if [ "$CHECK_COUNT" -gt 0 ] && [ "$(_count_pending "$CHECKS")" -eq 0 ]; then
+    break
+  fi
+  if [ "$CI_ELAPSED" -ge "$CI_TIMEOUT" ]; then
+    echo "PR #$PR_NUMBER: CI did not settle within ${CI_TIMEOUT}s. Still pending:"
+    echo "$CHECKS" | python3 -c "import json,sys
+for c in json.load(sys.stdin):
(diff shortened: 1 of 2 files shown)


VytCepas:
**Review response (Copilot):** three points, two valid:
- **`monitor_pr.sh:166` IndentationError — not reproduced (false positive).** The embedded `python3 -c` snippet runs correctly: `import`/`for` are at column 0 and `if`/`print` indented 4/8 (verified with `cat -A` and by executing it — it prints the pending check names). No change.
- **`:158` the repo's *root* `.claude/scripts/monitor_pr.sh` is still unbounded — correct.** That's the scope of **#222 (PI-203)**, which resyncs the root copy; I'll ensure the bounded CI-wait loop lands there (this PR fixes the scaffolded *template*).
- **`:191` weak contract test — fixed.** It now asserts the actual guard `[ "$CI_ELAPSED" -ge "$CI_TIMEOUT" ]` and the `CI_ELAPSED=$((CI_ELAPSED + …))` increment, so it fails if the loop goes unbounded again while the declarations linger. Test passes, ruff clean.

Issue #179, opened 2026-06-14
Initial scaffold silently overwrites existing CLAUDE.md and settings.json despite docs promising preservation


VytCepas:
## Summary

Running `project-init` inside a project that **already has** a hand-written `CLAUDE.md` and/or `.claude/settings.json` **silently overwrites them** — no warning, no backup, no `.new` sibling. Only `.claude/memory/` and `.claude/vault/` are preserved (`_PRESERVE_DIRS = {"memory", "vault"}` in `src/project_init/scaffold.py`). This contradicts the README, which tells users it is safe to re-run.

> README.md:145 — "Re-run any time — it reconciles, **preserves existing project files**, and never overwrites memory or vault notes."
> docs/development/template-system.md:46 — "Re-running the wizard **merges** scaffold files into the target project."

"Preserves existing project files" / "merges" overpromises: for everything the scaffolder ships a template for (`CLAUDE.md`, `.claude/settings.json`, `AGENTS.md`, `GEMINI.md`, `justfile`, …) the engine does a full overwrite. The conflict-aware `<file>.new`-sibling protection described at README.md:204 lives only in the `project-init upgrade` path (`src/project_init/upgrade.py`), **not** in the initial scaffold path (`scaffold.py::scaffold`).

This is the very common case: most projects adopting Claude Code already have a `CLAUDE.md` before they discover project-init.

## Affected configs

All presets / languages — it is in the core scaffold engine, independent of flags.

## Steps to reproduce (v0.3.0 wheel)

```bash
WHEEL=dist/project_init-0.3.0-py3-none-any.whl   # built via `uv build`
mkdir -p /tmp/probe && cd /tmp/probe
printf '# MY CUSTOM CLAUDE INSTRUCTIONS - DO NOT LOSE\n' > CLAUDE.md
mkdir -p .claude && printf '{"MY":"PRECIOUS SETTINGS"}\n' > .claude/settings.json
git init -q && git add -A && git commit -qm init

uvx --from "$WHEEL" project-init . \
  --non-interactive --preset obsidian-only --name existing \
  --description "overwrite probe" --language python

grep -c "DO NOT LOSE" CLAUDE.md                 # -> 0  (user content gone)
grep -c "PRECIOUS SETTINGS" .claude/settings.json  # -> 0  (user content gone)
ls CLAUDE.md*                                   # -> only CLAUDE.md, no CLAUDE.md.new
```

## Expected

[...]

PR #218, opened 2026-06-17
fix(PI-196): key first-scaffold on the record marker, not config existence

### LINKED ISSUE #196: First scaffold into a project that already has .claude/config.yaml skips overwrite protection (data loss)
### What happened
`scaffold()` decides whether overwrite protection applies by testing only the *existence* of the sentinel file:

```python
# src/project_init/scaffold.py:305
first_scaffold = not (target / ".claude" / "config.yaml").exists()
```

Protection (writing a `.new` sibling instead of clobbering a user file) only runs when `first_scaffold` is true or an unresolved sibling is already pending (`_protected_as_sibling`, scaffold.py:159). So if a target already contains a `.claude/config.yaml` that was **not** written by a prior project-init run (hand-authored, or left by another tool), the very first `project-init` run is misclassified as a re-run and pre-existing user files are overwritten with no `.new` sibling.

This is a residual gap in the PI-179 fix (#179): that issue fixed the common case, but the "have I run here before?" signal is file existence rather than "does `config.yaml` contain the scaffold record marker."

### Steps to reproduce
1. In a project that has a hand-written `.claude/config.yaml` and a custom `CLAUDE.md`, run `project-init . --non-interactive --preset obsidian-only --name x --description y`.
2. `CLAUDE.md` is overwritten in place; no `CLAUDE.md.new` is produced.

### Expected behaviour
Treat the run as a first scaffold (protect existing files) unless `.claude/config.yaml` actually contains the scaffold **record marker** (`_RECORD_MARKER`) — i.e. base `first_scaffold` on a real prior project-init run, not on the filename existing.

### Area / component
src/project_init/scaffold.py (`scaffold` / `_protected_as_sibling`). Relates to closed #179.

_Found by an independent Codex (codex-cli) review pass; verified against current code by Claude. Agent ready: Yes._

---

## Planning metadata
- Type: bug
- Priority: medium
- Size: S
- Area: scaffold.py (engine)
- Agent ready: Yes
- Confidence: high

## References
- #190, #197

## Dependencies
- None

## Acceptance criteria
- first_scaffold is decided by the record marker not file existence, so pre-existing files are protected on a true first run.

## Definition of Ready

## Definition of Done

VytCepas:
## Summary

`scaffold()` decided `first_scaffold` from the mere **existence** of `.claude/config.yaml`. A project that already had a `config.yaml` **not** written by project-init (hand-authored, or left by another tool) was misclassified as a re-run, so the first-scaffold overwrite protection (PI-179) was skipped — a pre-existing `CLAUDE.md` / `settings.json` could be silently overwritten instead of preserved as a `.new` sibling. A residual data-loss gap in the #179 fix.

## Fix

Base `first_scaffold` on the presence of the scaffold **record marker** that `write_scaffold_record` appends — the real signature of a prior project-init run. Moves `_RECORD_MARKER` to `scaffold.py` (the import root; `upgrade` already imports from it) and re-imports it in `upgrade.py`, so there's now a single source of truth for the marker.

## Test

Adds `test_pre_existing_config_without_record_still_protects` — a `config.yaml` lacking the marker still yields `CLAUDE.md` preserved + `CLAUDE.md.new`.

Closes #196


VytCepas:
**Review response (Copilot):** Fixed — extracted `_has_scaffold_record()` which wraps the `config.yaml` read in `try/except OSError`. A file that exists but can't be read (permissions) or vanishes between the `exists()` check and the read is now treated as *unrecorded* → `first_scaffold=True`, so overwrite protection stays on and user files are preserved as `.new` siblings instead of aborting. Added `test_unreadable_config_marker_is_treated_as_first_scaffold` (monkeypatches the config read to raise). 11 tests pass, ruff clean (helper extraction also kept `scaffold()` under the complexity gate).

PR #220, opened 2026-06-17
fix(PI-198): block gh-api merges regardless of flag position

### LINKED ISSUE #198: DAG guard gh-api-merge rule is bypassed by flags before the endpoint (gh api --method PUT ...)
### What happened
The PreToolUse guard rule that blocks PR merges via `gh api` requires the endpoint to come *immediately* after `gh api`:

```python
# templates/base/dot_claude/hooks/dag_workflow.py:247
re.compile(r"\bgh\s+api\s+repos/[^/\s]+/[^/\s]+/pulls/\d+/merge\b")
```

But merging via `gh api` requires a `PUT`, and the flag is naturally written **before** the endpoint:

```
gh api --method PUT repos/o/r/pulls/42/merge      # or:  gh api -X PUT repos/...
```

Here `--method` follows `gh api`, so `gh\s+api\s+repos/` does not match and the command falls through unguarded — bypassing the `monitor_pr.sh --merge` redirect that exists so CI/review gates are honored. The only form the rule catches is the endpoint-first `gh api repos/.../merge --method PUT`.

### Steps to reproduce
1. With the guard active, run `gh api --method PUT repos/<o>/<r>/pulls/<n>/merge`.
2. The guard does not block it; the merge proceeds.

### Expected behaviour
Match a gh-api merge regardless of flag position — detect `gh api` and `repos/.../pulls/\d+/merge` independently within the command so flags between them don't defeat the rule.

### Area / component
templates/base/dot_claude/hooks/dag_workflow.py (`COMMAND_RULES`); mirror the fix in the `plugins/` and repo `.claude/` copies. Defense-in-depth (git hooks + branch protection remain the real gate), but the guard currently gives false assurance.

_Found by an independent Codex review pass; verified against current code by Claude. Agent ready: Yes._

---

## Planning metadata
- Type: bug
- Priority: medium
- Size: XS
- Area: dag_workflow.py guard (template + copies)
- Agent ready: Yes
- Confidence: high

## References
- #202

## Dependencies
- None

## Acceptance criteria
- gh api merges are blocked regardless of flag position.

## Definition of Ready

## Definition of Done

VytCepas:
## Summary

The PreToolUse guard rule that blocks PR merges via `gh api` required the endpoint *immediately* after `gh api`:
`\bgh\s+api\s+repos/.../pulls/\d+/merge\b`. But merging via gh api needs a `PUT`, and the flag is naturally written **before** the endpoint — `gh api --method PUT repos/o/r/pulls/42/merge` did **not** match and bypassed the `monitor_pr.sh --merge` redirect that honors CI/review gates.

## Fix

Allow flags between `gh api` and the endpoint with `[^&|;]*?` (non-greedy, and stops at a command separator so it can't false-match across `&&`/`;`). The endpoint-first form still matches.

Mirrored to all three copies (template, repo `.claude/`, and the plugin payload via `tools/sync_plugin.py`).

## Test

Adds `test_github_command_guard_blocks_gh_api_merge_with_flags` — runs the scaffolded guard hook on `gh api --method PUT …/merge` and asserts `block`.

Closes #198

### DIFF
--- a/.claude/hooks/dag_workflow.py
+++ b/.claude/hooks/dag_workflow.py
@@ -244,7 +244,10 @@ COMMAND_RULES: list[tuple[re.Pattern[str], str | None, str]] = [
         "Direct pushes to main/master are blocked. Open a feature branch and PR.",
     ),
     (
-        re.compile(r"\bgh\s+api\s+repos/[^/\s]+/[^/\s]+/pulls/\d+/merge\b"),
+        # `[^&|;]*?` allows flags between `gh api` and the endpoint (a real merge
+        # needs `--method PUT`/`-X PUT`, often written before the path) without
+        # crossing a command separator (PI-198).
+        re.compile(r"\bgh\s+api\b[^&|;]*?repos/[^/\s]+/[^/\s]+/pulls/\d+/merge\b"),
         "pr.merged",
         "Use .claude/scripts/monitor_pr.sh <pr> --merge instead of `gh api .../merge` so CI and review gates are honored.",
     ),
--- a/.claude/hooks/dag_workflow.py
+++ b/.claude/hooks/dag_workflow.py
@@ -244,10 +244,11 @@ COMMAND_RULES: list[tuple[re.Pattern[str], str | None, str]] = [
         "Direct pushes to main/master are blocked. Open a feature branch and PR.",
     ),
     (
-        # `[^&|;]*?` allows flags between `gh api` and the endpoint (a real merge
-        # needs `--method PUT`/`-X PUT`, often written before the path) without
-        # crossing a command separator (PI-198).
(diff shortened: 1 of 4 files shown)


VytCepas:
**Review response (Copilot):** Fixed — the exclusion segment stopped at the first literal pipe, so a `gh api` call with a `--jq` pipe filter before the endpoint never reached the merge path and bypassed the guard. Dropped the pipe from the exclusion set (now `[^&;]*?`), keeping `;`/`&` as command separators — the endpoint structure itself is rigid (`repos / owner / repo / pulls / N / merge`), so this can't introduce false positives. Applied to all three `dag_workflow.py` copies (repo, plugin, template — sync test green) plus a new `--jq`-pipe regression test. 45 guard tests pass. (Amusingly, the stricter guard then blocked my own commit command for containing the sample endpoint string — confirmation it works.)

PR #225, opened 2026-06-17
fix(PI-188): reject path traversal in load_preset

### LINKED ISSUE #188: load_preset interpolates --preset into a filesystem path with no traversal guard
### What happened
`scaffold.load_preset(name)` builds a filesystem path directly from the `--preset` value with no validation:

```python
# src/project_init/scaffold.py:49
path = _TEMPLATES_DIR / "presets" / f"{name}.toml"
```

In interactive mode the preset comes from a fixed menu, but `--non-interactive --preset <value>` passes the value straight through. A value containing path separators or `..` (e.g. `--preset ../../../tmp/evil`) resolves outside `presets/` and would attempt to load an arbitrary `.toml`. Low severity (local CLI, user supplies their own input), but it is free hardening.

### Expected behaviour
Reject preset names that are not a bare stem: validate against the known preset list (`list_presets()`), or assert there are no path separators / `..` before constructing the path.

### Area / component
src/project_init/scaffold.py (`load_preset`).

---

## Planning metadata
- Type: bug
- Priority: low
- Size: XS
- Area: scaffold.py load_preset (engine)
- Agent ready: Yes
- Confidence: high

## References
- None

## Dependencies
- None

## Acceptance criteria
- The preset input rejects path separators and parent-dir traversal and is validated against known presets.

## Definition of Ready

## Definition of Done

VytCepas:
## Summary

`load_preset()` built `presets/<name>.toml` directly from the `--preset` value with no validation. In interactive mode the value comes from a fixed menu, but `--non-interactive --preset <value>` passes it straight through — `--preset ../../../tmp/evil` resolved outside `presets/` and would read an arbitrary `.toml`. Low severity (local CLI, user's own input), but free hardening.

## Fix

Reject names that aren't a bare stem (path separators, `..`, empty) **and** verify the resolved path's parent is still `presets/`; otherwise raise the existing `Unknown preset` error. Valid names and the existing not-found behavior are unchanged.

## Test

Adds a parametrized `test_load_preset_rejects_path_traversal` over `../../etc/passwd`, `..`, `a/b`, `foo/../bar`, `/etc/passwd`, `""`.

Closes #188

### DIFF
--- a/src/project_init/scaffold.py
+++ b/src/project_init/scaffold.py
@@ -48,9 +48,13 @@ def list_presets() -> list[dict]:
 
 def load_preset(name: str) -> dict:
     """Load a single preset by name (e.g. 'obsidian-only')."""
-    path = _TEMPLATES_DIR / "presets" / f"{name}.toml"
-    if not path.exists():
-        available = [p.stem for p in (_TEMPLATES_DIR / "presets").glob("*.toml")]
+    presets_dir = _TEMPLATES_DIR / "presets"
+    path = presets_dir / f"{name}.toml"
+    # A preset name is a bare stem, never a path — reject traversal so a value
+    # like `--preset ../../x` can't read a .toml outside presets/ (PI-188).
+    safe_name = bool(name) and not ({"/", "\\"} & set(name)) and ".." not in name
+    if not safe_name or path.resolve().parent != presets_dir.resolve() or not path.exists():
+        available = [p.stem for p in sorted(presets_dir.glob("*.toml"))]
         msg = f"Unknown preset {name!r}. Available: {', '.join(available)}"
         raise ValueError(msg)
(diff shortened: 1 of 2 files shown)


VytCepas:
**Review response (Copilot):** Fixed — added Windows-style separator cases to the parametrized `test_load_preset_rejects_path_traversal` in `tests/unit/test_presets.py`: `a\\b` and `..\\..\\etc\\passwd`. The `load_preset` guard already rejects backslash as a path separator (the `{"/", "\\"} & set(name)` check), so both new cases exercise that branch and pass. Verified locally: `uv run pytest tests/unit/test_presets.py -q` → 12 passed; `uv run ruff check tests/unit/test_presets.py` clean. Pushed as commit 11e34d5.

PR #231, opened 2026-06-17
test(PI-195): cover the interactive wizard leaf parsers and install.sh

### LINKED ISSUE #195: No execution coverage for interactive-wizard leaf parsers and install.sh
### What needs doing
Two areas of real logic have zero execution coverage:

**1. Interactive-wizard leaf parsers.** The orchestrator `_gather_inputs_interactive` is covered via stubs, but every leaf is monkeypatched away in the suite, so these branches never execute:
- `_choose_mcps_interactive` number-parse + dedup (`src/project_init/__main__.py:175-184`)
- `_choose_preset_interactive` out-of-range fallback (`__main__.py:152-154`)
- `_choose_db_interactive` mapping (`__main__.py:199-203`)

This is the only substantial untested source logic. Add unit tests that feed the parsers raw input strings directly.

**2. install.sh.** The 125-line `curl | bash` bootstrap is only string-checked (`test_release_engineering.py` reads it as text), never executed. Add an integration test that runs it against a temp `HOME`/target (mocking network/uv as needed) to catch shell regressions in the primary install path.

### Why now
These are the user's first-contact paths (the installer) and the wizard's actual input handling — the parts most likely to break unnoticed, and the suite's largest genuine gap.

### Area / component
tests (unit + integration); src/project_init/__main__.py, install.sh.

---

## Planning metadata
- Type: test
- Priority: medium
- Size: M
- Area: tests
- Agent ready: Yes
- Confidence: high

## References
- None

## Dependencies
- None

## Acceptance criteria
- Interactive wizard leaf parsers and install.sh have execution coverage.

## Definition of Ready

## Definition of Done

VytCepas:
## Summary
Two areas had zero execution coverage:
- **Wizard leaf parsers** (`_choose_preset_interactive`, `_choose_mcps_interactive`, `_choose_db_interactive`) — the integration tests monkeypatched every leaf away, so their number-parse/dedup/out-of-range branches never ran.
- **install.sh** — the 125-line curl|bash bootstrap was only read as text, never run.

## Tests added
- `tests/unit/test_interactive_prompts.py` — drives each leaf parser directly with canned prompts (out-of-range fallback, comma-parse + dedup, empty-skip, db mapping).
- `tests/integration/test_install_script.py` — `bash -n` syntax gate + an end-to-end run with **stubbed uv/git** and `PROJECT_INIT_REF=main` (no network/installs), asserting the `/project-init` slash command is written.

Both are new files (no overlap with other open PRs).

Closes #195

### DIFF
--- a/tests/integration/test_install_script.py
+++ b/tests/integration/test_install_script.py
@@ -28,10 +28,18 @@ def test_install_sh_writes_slash_command_with_stubs(tmp_path: Path):
     bindir = tmp_path / "bin"
     bindir.mkdir()
     # Stub the external tools so the bootstrap runs without network or installs.
-    for tool in ("uv", "git"):
-        stub = bindir / tool
-        stub.write_text("#!/usr/bin/env bash\nexit 0\n")
-        stub.chmod(0o755)
+    # uv is a pure no-op; git minimally emulates `clone <url> <dest>` by creating
+    # <dest>/.git, so install.sh runs against a real clone directory instead of an
+    # all-no-op stub that hides whether the clone path actually ran (PI-195 review).
+    (bindir / "uv").write_text("#!/usr/bin/env bash\nexit 0\n")
+    (bindir / "uv").chmod(0o755)
+    git_stub = bindir / "git"
+    git_stub.write_text(
+        "#!/usr/bin/env bash\n"
+        'if [ "$1" = clone ]; then mkdir -p "${@: -1}/.git"; fi\n'
+        "exit 0\n"
+    )
+    git_stub.chmod(0o755)
 
     env = {
         "HOME": str(home),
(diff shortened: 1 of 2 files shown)


VytCepas:
**Review response (Copilot):** Partly correcting, partly applying. The stub was `exit 0` for *every* git call, so `git -C` against a missing dir never actually failed under `set -e` — the test passed correctly. But the underlying point is fair: an all-no-op stub doesn't prove the clone path ran. **Fixed** by emulating `git clone <url> <dest>` (creates `<dest>/.git`) and adding `assert INSTALL_DIR/.git exists`, so the clone branch is now genuinely exercised. Tests pass, ruff clean.

PR #233, opened 2026-06-17
fix(PI-187): parse multi-line YAML safety.allow lists, not just inline JSON

### LINKED ISSUE #187: prod_guard.py safety.allow ignores multi-line YAML lists (escape hatch silently fails)
### What happened
The documented `safety.allow` escape hatch in `prod_guard.py` only parses an **inline JSON** list. The parser (`templates/fallback/dot_claude/hooks/prod_guard.py:78-90`) does:

```python
if stripped.startswith("allow:"):
    raw = stripped.split(":", 1)[1].strip()
    return [re.compile(p) for p in json.loads(raw)]
```

A user who writes the natural multi-line YAML form:

```yaml
safety:
  allow:
    - "git push --force-with-lease"
```

gets `raw == ""` → `json.loads("")` raises `JSONDecodeError` → swallowed by the `except` → returns `[]`. The allowlist silently does nothing and the user's commands keep getting blocked, with no indication the config failed to parse.

### Steps to reproduce
1. In a scaffolded project, add a multi-line `safety.allow` YAML list to `.claude/config.yaml`.
2. Run a command that the entry should allow.
3. It is still blocked; nothing warns that the allowlist did not parse.

### Expected behaviour
Either parse multi-line YAML lists (collect subsequent `- "…"` lines), or warn to stderr when `safety.allow` is present but unparseable. It fails safe today, but the feature is effectively non-functional in its most natural form.

### Area / component
templates/fallback hooks (`prod_guard.py`); mirror fix in `plugins/project-init-workflow/hooks/prod_guard.py`.

---

## Planning metadata
- Type: bug
- Priority: medium
- Size: S
- Area: prod_guard.py (templates/fallback + plugins)
- Agent ready: Yes
- Confidence: high

## References
- None

## Dependencies
- None

## Acceptance criteria
- Multi-line YAML safety.allow lists are parsed, or an unparseable allow warns to stderr.

## Definition of Ready

## Definition of Done

VytCepas:
## Summary

The documented `safety.allow` escape hatch in `prod_guard.py` only parsed an **inline JSON** list — `raw = allow:` remainder, then `json.loads(raw)`. A user writing the natural multi-line YAML form:

```yaml
safety:
  allow:
    - "kubectl delete .* --context kind-dev"
```

got `raw == ""` → `json.loads("")` raised → caught → `[]`. The allowlist silently did nothing, and the user's allowed commands kept getting blocked with no indication the config failed to parse.

## Fix

Parse **both** forms: inline JSON (`allow: ["a", "b"]`) and a multi-line YAML list (`allow:` followed by `- "a"` items). Fail-open behavior (a corrupt/unparseable allow → `[]`, guard still active) is unchanged. Mirrored to the plugin copy.

## Test

Adds `test_allowlist_multiline_yaml_suppresses_flag` (multi-line allow suppresses the flagged command; a non-listed command is still blocked). Existing inline + corrupt-fail-open tests still pass.

Closes #187

### DIFF
--- a/plugins/project-init-workflow/hooks/prod_guard.py
+++ b/plugins/project-init-workflow/hooks/prod_guard.py
@@ -65,7 +65,11 @@ def _find_config(start: Path) -> Path | None:
 
 
 def _allow_patterns(root: Path) -> list[re.Pattern[str]]:
-    """Read the safety.allow JSON list from .claude/config.yaml (fail-open).
+    """Read the safety.allow list from .claude/config.yaml (fail-open).
+
+    Accepts both an inline JSON list (``allow: ["a", "b"]``) and a multi-line
+    YAML list (``allow:`` on its own line followed by ``- "a"`` items). The
+    inline-only parser silently dropped the natural YAML form to ``[]`` (PI-187).
 
     *root* is the Bash tool's cwd, which may be a subdirectory after `cd` —
     the config is located by walking up the tree.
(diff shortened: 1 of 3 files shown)


VytCepas:
- `prod_guard.py:104` inline-JSON `allow` — **Fixed.** A scalar (JSON string/object) is no longer iterated character-by-character; we only `extend` from a list of strings, so a misconfigured `allow` can't silently disable the guard. Added regression test `test_scalar_inline_allow_does_not_overpermit`.
- `prod_guard.py:97` quote stripping — **Fixed.** Replaced `strip("\"'")` with `_unquote()`, which only removes a *matched* surrounding quote pair, leaving mismatched/single quotes intact.
- Applied identically to both copies (`templates/fallback` + `plugins/project-init-workflow`); 52 prod_guard tests pass, ruff clean.

PR #238, opened 2026-06-17
feat(PI-214): derive scaffolded CI Python matrix from requires-python

### LINKED ISSUE #214: Derive a Python version matrix for scaffolded CI from requires-python
## Summary

Follow-up to #208. That fix removed the hardcoded Python version so `uv sync` resolves the project's interpreter — CI now tests a **single** version (whatever the project pins). The *enhancement* half of #208 remains: test across **multiple** supported Python versions via a matrix, so version-specific dependency-resolution problems surface early.

The matrix must respect `requires-python` — a fixed matrix would reintroduce the #208 drift. The robust design derives the version set from the project's `requires-python` at scaffold time, which needs a new scaffold variable (e.g. `{{python_versions}}`) collected/derived by the wizard and rendered into `ci.yml.tmpl`. The template already carries a comment marking where the matrix slots in.

## Metadata
- Type: feature
- Priority: low
- Size: M
- Area: ci.yml.tmpl + wizard variable plumbing
- Agent ready: No (needs a design decision on how versions are collected/derived)
- Confidence: medium

## References
- #208

## Dependencies
- None

## Acceptance criteria
- Scaffolded ci.yml runs a Python matrix whose floor respects requires-python and cannot drift from it.

## Definition of Ready
- [ ] Decide how the version set is obtained (parse requires-python, ask in the wizard, or a sensible default)

## Definition of Done
- [ ] Matrix renders correctly for language=python scaffolds
- [ ] Regression test asserts the matrix respects the requires-python floor

VytCepas:
## Summary

Follow-up to #208/#209. The single-version CI fix landed; this adds the **multi-version matrix** half — and does it so the version floor can never drift from the project's own `requires-python` (the failure mode #208 warned a hardcoded matrix would reintroduce).

## Approach — runtime derivation (Option C)

A new **`python-matrix`** CI job reads `requires-python` from the project's `pyproject.toml` at workflow time and emits every supported CPython that satisfies it (filtered with `packaging.SpecifierSet`). `lint-and-test` then `needs:` that job and fans out via `strategy.matrix` over `fromJSON(needs.python-matrix.outputs.versions)`, pinning each leg with `python-version: ${{ matrix.python-version }}`.

- **Floor is always read from `requires-python`** → cannot drift. Only the *ceiling* enumeration (`KNOWN = 3.11–3.14`) is a maintained list; bump it (or let Renovate) when a new CPython ships.
- **No scaffold variable needed** — pure CI YAML, so the scaffolder stays untouched (the deliberately-noted fact that the scaffolder generates no `pyproject.toml` and never reads `requires-python` is *why* scaffold-time derivation was rejected).
- **Python-only** via `{{#if python}}`; node/go scaffolds are unchanged (verified: their `ci.yml` carries no `python-matrix` job).
- **Fresh-scaffold safe**: missing/unparseable `requires-python` → full `KNOWN` set; a `requires-python` that excludes every `KNOWN` → newest, so CI never errors on an empty matrix.

## Acceptance criteria / DoR / DoD

- ✅ Scaffolded `ci.yml` runs a Python matrix whose floor respects `requires-python` and cannot drift from it.
- ✅ DoR decision (how the version set is obtained): **derive at CI runtime from `requires-python`** — the only option that satisfies "cannot drift."
- ✅ Matrix renders correctly for `language=python` scaffolds; absent for node/go.
- ✅ Regression test asserts the matrix respects the `requires-python` floor — the test **extracts and executes the actual shipped resolver** from the rendered `ci.yml` against synthetic `requires-python` values (`>=3.13` → `["3.13","3.14"]`, `>=3.11,<3.13` → `["3.11","3.12"]`, missing → full set, `>=3.99` → newest).

## Test plan

- `uv run pytest` — 611 passed (5 new behavioral resolver tests + structural + node-negative).
- Rendered `ci.yml` for python/node/go all parse as valid YAML; job topology confirmed (`python-matrix` only for python).
- `ruff check` clean.
- Existing PI-208 guard (`test_ci_does_not_hardcode_python_version`) still passes — `${{ matrix.* }}` is not a literal pin.

Closes #214

### DIFF
--- a/templates/base/dot_github/workflows/ci.yml.tmpl
+++ b/templates/base/dot_github/workflows/ci.yml.tmpl
@@ -13,9 +13,79 @@ permissions:
   pull-requests: write
 
 jobs:
+{{#if python}}
+  # PI-214: derive the test matrix from requires-python at runtime, so the
+  # version floor can never drift from what the project itself declares (a
+  # hardcoded matrix would reintroduce the PI-208 drift). This job reads
+  # requires-python and emits every supported version that satisfies it;
+  # lint-and-test then fans out across them. KNOWN is the only maintained
+  # list — the *ceiling* set enumerated up to; bump it (or let Renovate) when
+  # a new CPython is released. The floor is always taken from requires-python.
+  python-matrix:
+    name: Resolve Python matrix
+    runs-on: ubuntu-24.04
+    outputs:
+      versions: ${{ steps.resolve.outputs.versions }}
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v8.1.0
+        with:
+          version: "0.11.7"
+
+      - name: Resolve versions satisfying requires-python
+        id: resolve
+        run: |
+          uv run --no-project --with packaging python - <<'PY'
+          import json
+          import os
+          import pathlib
+          import tomllib
+
+          from packaging.specifiers import SpecifierSet
+
+          # Maintained ceiling set — every CPython this template is willing to
+          # test. The floor comes from requires-python (below), never from here.
+          KNOWN = ["3.11", "3.12", "3.13", "3.14"]
+
+          spec = ""
+          pyproject = pathlib.Path("pyproject.toml")
+          if pyproject.is_file():
+              try:
+                  data = tomllib.loads(pyproject.read_text(encoding="utf-8"))
+                  spec = data.get("project", {}).get("requires-python", "") or ""
+              except (tomllib.TOMLDecodeError, OSError):
+                  spec = ""
+
+          if spec:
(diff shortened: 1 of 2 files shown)


VytCepas:
- `ci.yml.tmpl:66` — `SpecifierSet(spec)` raising `InvalidSpecifier` on a malformed `requires-python`: **Fixing.** Correct catch — the `try/except` only wrapped the TOML parse, so a present-but-invalid specifier (valid TOML, invalid PEP 440) would crash the `python-matrix` job rather than fall back. Now wrapped `SpecifierSet(spec)` in `except InvalidSpecifier:` → falls back to the full KNOWN set, with a regression test (`requires-python = "not-a-version"`). Fixed in aa99ac9.

Issue #250, opened 2026-06-17
Upstream pull-and-recommend: version-span detection + optional download recommendations


VytCepas:
## Summary

Let a fork pull upstream and see "what's new" as opt-in recommendations rather than forced changes. Requires recording version history (today only the scaffold-time version is stored — add `version_prev`/span and the plugin version), then on upgrade computing what is newly offered upstream since the recorded version and presenting it as a recommendation list to download. Pairs with surfacing the changelog for the span (#244) and with upgrade-as-a-PR (#241) so recommendations can arrive as reviewable PRs.

## Metadata
- Type: feature
- Priority: medium
- Size: L
- Area: upgrade.py + config record + release/changelog plumbing
- Agent ready: No
- Confidence: medium

## References
- #244 (surface changelog at upgrade)
- #241 (upgrade-as-a-PR)
- ADR-013

## Dependencies
- Builds on the version/plugin-version recording from the fork/pin issue and the consent issue

## Acceptance criteria
- [ ] Scaffold record carries enough version history to compute "new since last upgrade".
- [ ] Upgrade surfaces a recommendation list of newly-available upstream additions (opt-in to download).
- [ ] Recommendations never apply automatically.

## Definition of Ready
- [ ] Decide the source of "what's new" at upgrade time (embedded manifest diff vs fetched changelog).

## Definition of Done
- [ ] Version span recorded + computed; recommendation list generated and presented; tests for a known span.

PR #288, opened 2026-06-18
feat(PI-249): opt-in consent for new additions during upgrade

### LINKED ISSUE #249: Opt-in consent for new additions during upgrade (accept/reject each new item)
## Summary

Make upstream updates ask before adding NEW things. The drift engine already separates `report.new` (files that did not exist at scaffold) from `report.changed`, so consent is a targeted hook in `run_upgrade` after `compute_drift`: present each new addition (new skill/hook/file) and let the user accept or reject before `apply_drift` writes it. Rejected items are simply not applied (and can be offered again next time).

## Metadata
- Type: feature
- Priority: medium
- Size: M
- Area: upgrade.py (run_upgrade + apply_drift skip-list)
- Agent ready: Yes
- Confidence: high

## References
- #245 (per-file interactive apply — composes with this)
- ADR-013

## Dependencies
- Composes with #245; conceptually under ADR-013

## Acceptance criteria
- [ ] On upgrade, new additions are presented for per-item accept/reject (interactive), or accepted in bulk via a flag.
- [ ] Rejected new items are not written; existing `changed`/`conflict` behavior is unchanged.
- [ ] Non-interactive default remains backward compatible.

## Definition of Ready
- [ ] None — mechanism confirmed by analysis (report.new is already separable).

## Definition of Done
- [ ] Consent flow implemented with a skip-list into apply_drift; tests drive accept/reject.

---

## Grill+Codex revision (2026-06-18)

- **Stable "addition groups"** replace path-based consent: each group has an ID + path membership + rationale + recorded accept/decline state (current `report.new` is file paths, not stable IDs).
- Add `--accept-new <id>` / `--decline-new <id>`; **fail `--apply`** until choices are recorded.
- **Declined IDs are stored and suppressed** on future `--apply` unless the upstream addition changes materially (no perpetual-failure loop).
- **Pulled forward:** ships right after ADR-013, in parallel with #247 — this consent is version-independent (built on `report.new`). Only **span/recommendation** detection (#250) depends on #248's version schema.

---

## Revision (2026-06-18) — three-profile model

The `standalone` profile (copied-in, owner-driven updates) is the primary consumer of upgrade
consent: `upgrade` **must notify of potential merge conflicts** on customized copied-in files
(ties to #240 three-way merge, #244 changelog). Addition-group consent (this ticket) applies
across all three profiles; `org` additionally gates new additions through the fork's
pull-and-recommend (#250).

VytCepas:
Closes #249

Add opt-in consent for new additions during `upgrade` (ADR-013).

- **Addition groups**: genuinely-new files (absent from the recorded manifest) are bucketed by feature/overlay (`devcontainer`, `claude-skills`, `docs`, … + `misc`) with a stable ID, rationale, and content hash.
- **Consent gate**: `upgrade --apply` refuses (exit 2) until every new group is accepted or declined, listing the groups + how to decide.
- **Flags**: `--accept-new <id|all>` applies a group; `--decline-new <id|all>` records it (in `updates.declined_additions` as `{id: hash}`) and suppresses it.
- **Suppression + material change**: declined groups are skipped on future applies; a declined group whose content hash changes resurfaces as undecided.
- **Restorations aren't gated**: a manifested file the user deleted is a restoration (already consented to) and is applied without a decision — only genuinely-new files need consent.
- Tests: `tests/integration/test_consent.py` (classify, gate, accept/decline, suppression, restoration). 3 existing upgrade tests opt past the gate with `--accept-new all` where they intentionally bulk-apply new files. Full suite **665 green**.

Unblocks #250 (pull-and-recommend), which builds on this consent gate.

### DIFF
--- a/src/project_init/__main__.py
+++ b/src/project_init/__main__.py
@@ -588,8 +588,31 @@ def _upgrade_main(argv: list[str]) -> int:
         action="store_true",
         help="Accepted for CLI symmetry — upgrade never prompts",
     )
+    p.add_argument(
+        "--accept-new",
+        action="append",
+        default=[],
+        metavar="ID",
+        help="Accept an addition group on --apply (repeatable; 'all' accepts every new group, #249)",
+    )
+    p.add_argument(
+        "--decline-new",
+        action="append",
+        default=[],
+        metavar="ID",
+        help=(
+            "Decline an addition group; recorded and suppressed on future "
+            "--apply unless it changes materially ('all' declines every new group)"
+        ),
+    )
     args = p.parse_args(argv)
-    return run_upgrade(Path(args.target).resolve(), apply=args.apply, no_plugin=args.no_plugin)
+    return run_upgrade(
+        Path(args.target).resolve(),
+        apply=args.apply,
+        no_plugin=args.no_plugin,
+        accept_new=args.accept_new,
+        decline_new=args.decline_new,
+    )
 
 
 def _build_variables(preset: dict, inputs: ScaffoldInputs) -> dict[str, str]:
--- a/src/project_init/upgrade.py
+++ b/src/project_init/upgrade.py
@@ -595,7 +595,11 @@ def _addition_groups(new_paths: list[Path], staging: Path) -> dict[str, dict]:
     for group in groups.values():
         digest = hashlib.sha256()
         for rel in group["paths"]:
-            digest.update((staging / rel).read_bytes())
+            # Length-delimit path + content so different per-file splits can't
+            # collide (e.g. "ab"+"c" vs "a"+"bc" would otherwise match).
+            content = (staging / rel).read_bytes()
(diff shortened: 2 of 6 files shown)


VytCepas:
**Review response (Copilot):** both fixed (good catches).

- *hash collision from byte concatenation* (`upgrade.py:599`): the per-file update now length-delimits path + content — `f"{rel}\0{len(content)}\0"` then the bytes — so different per-file splits (or same bytes under different names) can't collide. Added `test_group_hash_is_path_sensitive`.
- *`json.loads` breaks on an inline YAML comment* (`upgrade.py:618`): `_read_declined` now extracts just the `{...}` object via regex before parsing, so a hand-added `# comment` after the value no longer silently drops recorded declines. Added `test_read_declined_tolerates_inline_comment`.

Full suite 667 green.

PR #294, opened 2026-06-18
feat(PI-242): guard upgrade --apply on a dirty git tree + undo hint

### LINKED ISSUE #242: Guard upgrade --apply on a dirty git tree and print an undo hint
## Summary

`upgrade --apply` writes files with no awareness of git state. Before applying, check whether the target is a git repo with a clean working tree; if dirty, refuse (or require `--force`) and tell the user to commit/stash first. On success, print how to undo (`git checkout .` / review with `git diff`). Makes apply trivially reversible and removes the "what if it clobbers my uncommitted work" fear.

## Metadata
- Type: feature
- Priority: medium
- Size: S
- Area: upgrade.py (run_upgrade pre-apply check)
- Agent ready: Yes
- Confidence: high

## References
- #142 (upgrade system)
- #179 / #196 (overwrite-protection hardening)

## Dependencies
- None

## Acceptance criteria
- [ ] `--apply` on a dirty tree warns and refuses unless `--force`.
- [ ] Non-git targets still work (guard is best-effort, not a hard requirement).
- [ ] On apply, an undo hint is printed.

## Definition of Ready
- [ ] None — behavior is well understood.

## Definition of Done
- [ ] Guard + `--force` implemented; undo hint printed.
- [ ] Tests: dirty tree blocked, --force overrides, non-git target tolerated.

VytCepas:
## Summary

Guards `project-init upgrade --apply` on a dirty git work tree (upgrade-UX cluster). Applying onto uncommitted work intermixes the upgrade with the user's edits and makes it hard to review or revert; this ensures the upgrade always lands as a single reviewable, revertible diff.

## Behavior
- **Dirty tree + `--apply`** → blocked with a "commit or stash first" message; exits **code 3** (distinct from the addition-consent gate's `2`).
- **`--force`** (alias `--allow-dirty`) → bypass for power users (prints an intermixed-diff caveat). `--force` is the flag named in #242's acceptance criteria.
- **Clean apply** → prints a git-based undo hint (`git diff` to review, `git restore .` to discard) — valid precisely because the guard ensured the upgrade is the only uncommitted change.
- **Non-git target** → soft note, proceeds (doesn't hard-block).

Implemented as a CLI-layer precondition in `_upgrade_main`, leaving `run_upgrade` unchanged for programmatic callers.

## Tests
5 new integration tests (clean / dirty / `--force` / `--allow-dirty` alias / non-git). Full suite green; `ruff check` clean; verified end-to-end via the real CLI.

Closes #242

### DIFF
--- a/src/project_init/__main__.py
+++ b/src/project_init/__main__.py
@@ -523,10 +523,7 @@ _PROFILE_SUMMARY = {
         "advisory enforcement (external official marketplace still enabled — "
         "full no-egress is the org mode, #258)"
     ),
-    "org": (
-        "fork as source-of-truth, host-adaptive delivery, hard (server-side) "
-        "enforcement"
-    ),
+    "org": ("fork as source-of-truth, host-adaptive delivery, hard (server-side) enforcement"),
 }
 
 
@@ -559,7 +556,12 @@ _LANGUAGE_COMMANDS: dict[str, tuple[str, str, str]] = {
 
 def _upgrade_main(argv: list[str]) -> int:
     """Parse and run the `project-init upgrade` subcommand (PI-142)."""
-    from project_init.upgrade import run_upgrade
+    from project_init.upgrade import (
+        _enforce_clean_tree,
+        _git_worktree_status,
+        _print_undo_hint,
+        run_upgrade,
+    )
 
     p = argparse.ArgumentParser(
         prog="project-init upgrade",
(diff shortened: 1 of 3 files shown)


VytCepas:
- `--force` vs `--allow-dirty` (`upgrade.py:800`, `__main__.py:622`, `test_upgrade.py:105`): Fixing — verified #242's acceptance criteria names the override `--force`. It is now the primary flag (messages, help, and tests updated); `--allow-dirty` is kept as an alias with its own test, so nothing breaks.
- Non-git note wording (`upgrade.py:783`): Fixing — the note now reads "not a git repository, or git status could not be determined" and suggests initializing git first, since the None case covers both no-repo and git-unavailable.

All four addressed in a383019.

VytCepas:
**Review response (Codex):**
- Scope undo command to target (`upgrade.py:823`): Fixing — good catch. Because the guard checks the target subtree (`git -C <target> status -- .`), a clean subdir passes even with dirty repo siblings, so a bare `git restore .` could discard them. Printed restore/diff/commit hints are now scoped via `git -C <target>` (bare when the target is the cwd), with a regression test for the clean-subdir + dirty-sibling case. Addressed in 1a1a7e8.

Issue #359, opened 2026-06-21
OS-agnostic scaffolded output + non-CLI agent surface integration


VytCepas:
## Summary

Make the **scaffolded output** work seamlessly across operating systems (macOS, Linux, Windows) **and** across non-CLI / GUI agent surfaces (Claude Code Desktop, VS Code agent mode, Codex IDE, Cursor, Antigravity, …) — not just Linux + the Claude Code CLI.

This epic comes out of a portability audit (June 2026). Two findings drive it:

1. **The scaffolder engine is already OS-agnostic** (pure `pathlib`/stdlib, platform-neutral installers, no `brew`/`apt`/`winget`, WSL only in troubleshooting notes). ✅ No work needed there beyond two cheap hardening nits.
2. **The scaffolded *shell/hook output* is Linux-biased and actively breaks on macOS**, and native Windows (no WSL) is non-viable. This violates the requirement that a macOS user hit **zero friction**.

A second, related axis: the same "where does the shell run" fault line determines how well the scaffold integrates with **non-CLI agent surfaces**. The ecosystem converged in late-2025/2026 — hooks, `SKILL.md`, MCP, and `AGENTS.md` are now broadly supported — but the wiring differs per tool, and the bash/git enforcement layer is a **local-surface-only guarantee** (cloud sandboxes honor only repo-committed config).

### A — OS-agnostic scaffolded output (mostly fixes)
The audit's must-fix set. macOS correctness blockers are wired into **always-on** hooks (first commit, first session), so they are high priority.

- macOS bash-3.2 / BSD-coreutils breakage: `mapfile` (`pre_commit_gate.sh:31,52`), `sha256sum` (`session_setup.sh:16`)
- Interpreter portability: hardcoded `python3` in ~20 sites incl. 3 JSON hook configs (`settings.json.tmpl:52`, `codex/hooks.json.tmpl:9`, `gemini/hooks.json.tmpl:10`); `dag_workflow.py:477` execs a shebang `.sh` directly
- Hardening: `stat -c` in `audit/SKILL.md:38`, unguarded `jq` in `setup_github.sh`, missing `gh` guards, shebang inconsistency, `.gitattributes` `*.sh eol=lf`, engine `write_text(newline="\n")`
- Portability regression tests + macOS CI runner

### B — Non-CLI / GUI agent surface integration (lots to determine → spikes first)
[...]

PR #375, opened 2026-06-21
fix(PI-360): macOS bash-3.2/BSD breakage in always-on hooks (mapfile, sha256sum)

### LINKED ISSUE #360: Fix macOS bash-3.2/BSD breakage in always-on hooks (mapfile, sha256sum)
Part of #359 (workstream A — OS portability).

## Policy (governs this fix)
**Shell floor = bash 3.2** (the version every macOS ships as `/bin/bash`; see #359). Keep arrays, `[[ ]]`, process substitution, herestrings — all present in 3.2. Forbid 4.0+ builtins only (`mapfile`/`readarray`, `declare -A`, `${var^^}`). No rewrite to POSIX `sh`. Lint enforcement of the forbidden set is tracked in #363.

## Problem
Two GNU/bash-4 constructs in **always-on** hooks break on stock macOS (bash 3.2 + BSD coreutils) — hit on first commit / first session:

1. `templates/fallback/dot_claude/hooks/pre_commit_gate.sh:31,52` — `mapfile` is a bash-4 builtin.
2. `templates/fallback/dot_claude/hooks/session_setup.sh:16` — `sha256sum` is GNU-only (macOS has `shasum`). Worse: the script is `set -uo pipefail` **without** `-e`, so the failed pipe yields an empty `CURRENT` → stamp never matches → **every session silently re-bootstraps** (`uv sync`/`bun install`). The fix kills this latent perf bug too.

### Fix 1 — `mapfile` → portable read loop (both sites)
Process substitution `< <(...)` is required (keeps the loop in the current shell so the array survives; a `|` pipe subshells and loses it):
```bash
STAGED_PY=()
while IFS= read -r f; do [ -n "$f" ] && STAGED_PY+=("$f"); done \
  < <(git diff --cached --name-only --diff-filter=ACM 2>/dev/null | grep '\.py$' || true)
```
Same shape for `STAGED_JS` (line 52) with the `grep -E '\.(js|ts|jsx|tsx)$'` filter. Newline-delimited = exact parity with current `mapfile -t` behavior (handles spaces in filenames; newline-in-filename is the only unsupported case, same as today).

### Fix 2 — `sha256sum` → hasher fallback in `session_setup.sh`
Replace the `fingerprint()` body. `sha256sum`, `shasum -a 256`, and `cksum` all emit their value in field 1, so one `cut` covers all three; `cksum` is the POSIX always-present floor that makes the function total:
```bash
_hash() {
  if   command -v sha256sum >/dev/null 2>&1; then sha256sum
  elif command -v shasum    >/dev/null 2>&1; then shasum -a 256
  else cksum
  fi | cut -d' ' -f1
}

fingerprint() {
  cat pyproject.toml uv.lock package.json bun.lock bun.lockb go.mod go.sum 2>/dev/null | _hash
}
```

## Implementation steps
1. Edit the two files in `templates/fallback/dot_claude/hooks/` (canonical source).
2. `uv run python tools/sync_plugin.py` — regenerates the plugin copies under `plugins/project-init-workflow/hooks/` (CI verifies they're in sync).
3. Add a focused temp-dir scaffold test (CLAUDE.md requires one for any `templates/` change): scaffold, then assert the hooks contain no `mapfile`/`sha256sum` and that `fingerprint`/the staged-file loop run under bash. (Broad portability lint = #363.)

## Acceptance
- `pre_commit_gate.sh` and `session_setup.sh` run clean on stock macOS bash 3.2; no `mapfile`/`sha256sum` anywhere on these paths.
- Plugin copies regenerated and in sync.
- Side effect: empty-hash re-bootstrap loop on macOS is gone.

## Scope notes
- Shebang standardization left to #362 (the 3.2-safe fixes here don't require it).

VytCepas:
## What

Workstream A of epic #359 (OS-agnostic scaffolded output). Fixes the two macOS correctness blockers wired into **always-on** hooks, so a macOS user (stock `/bin/bash` = bash 3.2, BSD coreutils) hits zero friction on first commit / first session.

- **`pre_commit_gate.sh`** — `mapfile` is a bash-4+ builtin absent from macOS bash 3.2. Replaced both `mapfile -t STAGED_PY` / `mapfile -t STAGED_JS` with a portable `while IFS= read -r` loop that builds the arrays; expanded arrays stay quoted.
- **`session_setup.sh`** — the bootstrap fingerprint used GNU `sha256sum`, which macOS does not ship. Now probes for `sha256sum` and falls back to BSD `shasum -a 256`.
- Synced the derived plugin payload via `tools/sync_plugin.py`.

## Shell policy

Floor = bash 3.2 (universal macOS bash). No POSIX-sh rewrite; only bash-4+ builtins forbidden.

## Tests

- New `tests/contracts/test_hook_portability.py` — asserts no bash-4 builtins in the always-on hooks (both fallback source + plugin payload), shasum-aware fingerprint, portable array build.
- Updated `test_scaffold_obsidian.py` gate assertion to the portable construct.
- Full suite: 845 passed, 3 skipped. `bash -n` syntax-clean; functional checks confirm equivalent behavior.

Part of #359.
Closes #360

### DIFF
--- a/plugins/project-init-workflow/hooks/pre_commit_gate.sh
+++ b/plugins/project-init-workflow/hooks/pre_commit_gate.sh
@@ -28,7 +28,11 @@ ERRORS=""
 # Lint and auto-fix staged Python files. Prefer 'uv run ruff' inside a
 # uv-managed project so the hook uses the same ruff the project itself uses;
 # fall back to a system ruff binary.
-mapfile -t STAGED_PY < <(git diff --cached --name-only --diff-filter=ACM 2>/dev/null | grep '\.py$' || true)
+# bash 3.2 (macOS /bin/bash) has no `mapfile`/`readarray` — read into the array
+# with a portable while-loop so the always-on commit gate runs everywhere.
+STAGED_PY=()
+while IFS= read -r _f; do [ -n "$_f" ] && STAGED_PY+=("$_f"); done \
+    < <(git diff --cached --name-only --diff-filter=ACM 2>/dev/null | grep '\.py$' || true)
 if [ "${#STAGED_PY[@]}" -gt 0 ]; then
     LINT_OUT=""
     if { [ -f "$ROOT/pyproject.toml" ] || [ -f "$ROOT/uv.lock" ]; } && command -v uv &>/dev/null; then
@@ -49,7 +53,9 @@ fi
 
 # Lint and auto-fix staged JS/TS files
 # Use bunx (bun's package runner) — consistent with project convention (PI-15).
-mapfile -t STAGED_JS < <(git diff --cached --name-only --diff-filter=ACM 2>/dev/null | grep -E '\.(js|ts|jsx|tsx)$' || true)
+STAGED_JS=()
+while IFS= read -r _f; do [ -n "$_f" ] && STAGED_JS+=("$_f"); done \
+    < <(git diff --cached --name-only --diff-filter=ACM 2>/dev/null | grep -E '\.(js|ts|jsx|tsx)$' || true)
 if [ "${#STAGED_JS[@]}" -gt 0 ] && command -v bunx &>/dev/null; then
     bunx eslint --fix --quiet "${STAGED_JS[@]}" 2>/dev/null || true
     LINT_OUT=$(bunx eslint --quiet "${STAGED_JS[@]}" 2>&1 || true)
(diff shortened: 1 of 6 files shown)


VytCepas:
- `session_setup.sh` hasher could go undefined on a host with neither `sha256sum` nor `shasum`: **Fixing** — added explicit `command -v shasum` probe and a final POSIX `cksum` fallback so `fingerprint()` is always defined (an empty digest would force a re-bootstrap every session). Applied to both fallback source and synced plugin copy.
- `test_hook_portability.py` didn't enforce the probe/fallback: **Fixing** — the test now asserts the hook probes for both `sha256sum` and `shasum` and provides the `cksum` fallback, so an unconditional `shasum` call would fail the suite.

PR #376, opened 2026-06-21
fix(PI-361): route all Python invocations through a canonical _py.sh resolver

### LINKED ISSUE #361: Resolve interpreter portability (python3 vs python) across scripts, JSON hook configs, and dag_workflow exec
Part of #359 (workstream A — OS portability). **Decision: (B) interpreter-agnostic, via one canonical resolver.** Bash 3.2 floor (#359).

## Why / scope
`python3` exists on macOS, Linux, and WSL — so this is **not** a Mac blocker (#360 is). It extends support to **native Windows (Git Bash, no WSL)** where only `python`/`py` may exist, and hardens the **uv-only-Python** edge (no system `python3`). Hooks are stdlib-only — any Python 3 works; they just need a *resolvable* interpreter.

## Approach: a single `_py.sh` wrapper, used everywhere
Centralize interpreter resolution in ONE file rather than scattering `command -v` logic across scripts. Add `templates/fallback/dot_claude/hooks/_py.sh` (synced to the plugin automatically — `sync_plugin.py` globs `*.sh`):
```bash
#!/usr/bin/env bash
PY="$(command -v python3 || command -v python || true)"
[ -z "$PY" ] && exec uv run python "$@"
exec "$PY" "$@"
```
It forwards `"$@"`, so it handles file form (`_py.sh script.py`), inline (`_py.sh -c "…"`), heredoc (`_py.sh - <<'PY'`), and env-prefixed (`VAR=x _py.sh -c …`) calls identically. Precedent: `templates/multi_model/.../setup_models.sh:109` already uses a `have python3 / uv run python` fallback.

## Sites — route every `python3` invocation through `_py.sh`
Reference it via `"$(dirname "$0")/_py.sh"` (hooks) or `"$(dirname "$0")/../hooks/_py.sh"` (scripts), mirroring how shims already locate `../hooks/dag_workflow.py`.

1. **`exec python3` shims:** `create_nojira_pr.sh:3`, `finish_pr.sh:3`, `promote_review.sh:3`, `push_branch.sh:5`, `github_command_guard.sh:11`.
2. **Inline `python3 -c` / heredocs:** `pre_commit_gate.sh:10,73` (coordinate with #360), `post_edit_lint.sh:12,51`, `workflow_state_reminder.sh:12,14` (keep the `DAG_STATE=…` env prefix before `_py.sh`), `monitor_pr.sh:65,75,170,176`, `create_issue.sh:455,469,504,520`.
3. **JSON hook configs** (already in this plan): `settings.json.tmpl:52`, `codex/dot_codex/hooks.json.tmpl:9`, `gemini/dot_gemini-extension/hooks/hooks.json.tmpl:10` → `_py.sh <script> [args]`. Also repoint the equivalent wiring in the **plugin** `hooks.json` if it calls `python3` directly (verify during impl).

## Other interpreter fixes
4. **`dag_workflow.py:477`** (not python resolution — shebang-exec of a `.sh`): replace `subprocess.run([".claude/scripts/monitor_pr.sh", …])` with an explicit bash call on an absolute path:
   ```python
   script = Path(__file__).resolve().parent.parent / "scripts" / "monitor_pr.sh"
   monitor_args = ["bash", str(script), str(pr_number), "--merge"]
   ```
5. **Python shebangs** (`dag_workflow.py:1`, `agent_guard_adapter.py.tmpl:1`): leave `#!/usr/bin/env python3` — correct on mac/linux/wsl, ignored on native Windows, no longer on the exec path (everything goes through `_py.sh`).
6. **Docs/skill guidance** (low priority, same ticket): `audit/SKILL.md`, `add_hook/SKILL.md`, `hooks/README.md` (×3 mirrors) — update "use python3" wording to reference `_py.sh`.

## Implementation steps
1. Add `_py.sh`; edit canonical sources (`templates/fallback/…` hooks, `templates/base/…` scripts/json, `dag_workflow.py`).
2. `uv run python tools/sync_plugin.py` (propagates `_py.sh` + edited hooks; verify plugin `hooks.json` wiring).
3. Temp-dir scaffold test: assert no `python3`/`python` invoked as a command anywhere except `_py.sh`; assert `_py.sh` resolves under a `python`-only PATH (shadow `python3`).

## Acceptance
- All Python invocations go through `_py.sh`; no bare `python3`/`python` command elsewhere.
- Hooks/scripts run on a `python`-only host (native Windows / Git Bash) and a uv-only host.
- Plugin copies regenerated and in sync. Lint rule enforced in #363.

VytCepas:
## What

Workstream A of epic #359. Implements the decided approach **(B) interpreter-agnostic via one canonical resolver**.

Hooks/scripts are stdlib-only, so any Python 3 works — but the command may be `python3` (mac/linux/wsl), `python` (native Windows / Git Bash), or only reachable via `uv run python` (uv-only host). This centralizes resolution in one file instead of scattering `command -v` logic.

## Changes

- **New `templates/fallback/dot_claude/hooks/_py.sh`** (auto-synced to the plugin): tries `python3` → `python` → `uv run python`; forwards `"$@"` so file / `-c` / heredoc / env-prefixed call forms all work.
- **Routed every `python3` invocation through it:**
  - exec shims: `push_branch.sh`, `finish_pr.sh`, `promote_review.sh`, `create_nojira_pr.sh`, `github_command_guard.sh`
  - inline `-c` / heredocs: `pre_commit_gate.sh`, `post_edit_lint.sh`, `workflow_state_reminder.sh`, `monitor_pr.sh`, `create_issue.sh`
  - `multi_model/setup_models.sh` cascade (kept the envsubst/cp fallback for no-Python hosts)
  - JSON hook configs: `settings.json.tmpl`, plugin `hooks.json`, `codex` + `gemini` `hooks.json.tmpl`
- **`dag_workflow.py` `cmd_finish`:** invoke `monitor_pr.sh` via explicit `bash` on an absolute path (`Path(__file__).resolve().parent.parent/…`) instead of exec'ing the `.sh` by relative path — no dependence on the exec bit, shebang resolution, or CWD.
- Updated `add_hook` / `audit` skills + `hooks/README.md` to teach the `_py.sh` pattern.
- Synced plugin payload and the repo-dogfooded `dag_workflow.py` (the only match-enforced dogfood copy).

## Tests

- New `tests/contracts/test_interpreter_portability.py`: asserts no bare `python3`/`python` command invocations outside `_py.sh`; resolver exists, is executable, has the uv fallback; JSON configs call `_py.sh`; `dag_workflow` uses `bash`+abspath; and a **functional check that `_py.sh` falls back to `python` under a `python`-only PATH**.
- Updated stale shim/marketplace/obsidian assertions for the new form.
- Full suite: **871 passed, 3 skipped**. End-to-end scaffold confirms `_py.sh` lands executable in `.claude/hooks/` in both plugin and `--no-plugin` modes.

Part of #359.
Closes #361


VytCepas:
**Review response (Codex):**
- **P1 — _py.sh missing in plugin-mode scaffolds: Fixing (confirmed real).** You're right: in true plugin mode the `fallback` layer isn't copied, so `.claude/hooks/` had only `README.md`+`dag_workflow.py` and the base lifecycle scripts (`push_branch.sh` etc.) referenced a non-existent `../hooks/_py.sh`. Fixed by moving `_py.sh` into `templates/base/dot_claude/hooks/` (always scaffolded, like `dag_workflow.py`) and teaching `sync_plugin.py` to mirror it into the plugin payload. Added a parametrized scaffold test (plugin **and** --no-plugin) asserting `_py.sh` lands executable in `.claude/hooks/` and the shim's referenced path resolves — this would have caught the bug.

VytCepas:
- **Codex P1 (resolver missing in plugin mode):** Already fixed in commit `80dd096` (the comment is on the older `e87d85a7`) — `_py.sh` moved to `templates/base/`, scaffolded in all modes, with a parametrized regression test.
- **Copilot — `_py.sh` uv fallback when uv absent / `python` may be Python 2:** **Fixing.** Rewrote the resolver: use `python3` directly (PEP 394 guarantees it's 3.x); accept bare `python` only if a version probe confirms ≥3; attempt `uv run python` only when `uv` exists; otherwise exit 127 with a clear `no Python 3 found` message. Applied to base + synced plugin copy.
- **Copilot — `setup_models.sh` runs the py3 generator whenever `python` exists (could be py2):** **Fixing.** The cascade now delegates the decision to `_py.sh` (which rejects Python 2) and falls back to `envsubst` then a literal copy if it can't run, so a py2-only host no longer skips a working `envsubst`.
- Added tests for the Python-2 rejection + clear-error path. Full suite: 874 passed.

Issue #365, opened 2026-06-21
Spike: validate non-CLI surface capability matrix (smoke-test scaffold on VS Code / Cursor / Codex IDE)


VytCepas:
Part of #359 (workstream B — non-CLI surfaces). **SPIKE / validation.**

Several capability-matrix claims are MEDIUM confidence and must be validated before building (#359 / B1):
- VS Code agent mode reads `.claude/` hooks + skills + CLAUDE.md directly (camelCase↔PascalCase conversion)?
- Cursor `hooks.json` event vocabulary + SKILL.md (v2.4) behavior?
- Codex IDE `.agents/skills` path + `hooks.json`/config.toml?
- Antigravity v2.0 `hooks.json` (docs are JS-rendered / un-fetchable).

**Deliverable:** smoke-test a scaffolded project on each surface; record actual behavior in a matrix that feeds B1's ADR.

PR #381, opened 2026-06-21
feat(PI-366): emit multi-surface MCP + hooks config from a canonical source

### LINKED ISSUE #366: Emit multi-surface MCP + hooks config from canonical source
Part of #359 (workstream B). **Implements #364 (Option 2 generator).** Depends on #364 (ADR) + #365 (validated wiring).

Build the generator that emits per-surface config from the canonical hook/MCP spec:
1. Define the **canonical specs** (hooks already exist in `templates/fallback`; add a canonical MCP spec).
2. **Surface table** (filename, MCP JSON key/shape, event-name casing, skills path) for Cursor / Codex IDE / Antigravity. (Claude + VS Code: no emission.)
3. **Generator** (extend the `tools/sync_plugin.py` pattern) that renders each in-scope surface's files.
4. **Opt-in**: wizard/flag for which surfaces to emit; record the choice in the `config.yaml` scaffold-record.
5. **Drift guard**: contract tests fail if a surface's generated config falls out of sync with the canonical spec.
6. Temp-dir scaffold tests per surface: assert each gets correct hook/MCP files; assert the expected shape (hook fires / MCP server discoverable).

Acceptance: enabling a surface emits working `hooks.json` + MCP config for that tool; the enforcement hooks/MCP become functional (and visible in that tool's panel); no drift vs the canonical source.

VytCepas:
## What

Implements **ADR-017** (epic #359 workstream B): one canonical source rendered outward to each non-CLI agent surface's native files. Per your direction — **all surfaces incl. Antigravity (experimental)**, **auto-emit tied to `--agents`** (no new flag).

## How

- **`src/project_init/surfaces.py`** — the canonical surface table + pure, deterministic renderers:
  - MCP: `mcpServers` JSON (Claude root `.mcp.json`, Cursor), VS Code `servers` JSON, Codex `[mcp_servers.*]` TOML.
  - Hooks: Cursor camelCase (`beforeShellExecution`/`beforeSubmitPrompt`), Antigravity `safety-gate`/`PreToolUse`.
  - `planned_files()` is the **single** selection→files map shared by the scaffold step and the drift test.
- **`mcps.py`** — structured `server` specs on catalog entries (the canonical MCP spec — MCP was never emitted as runnable config before) + `servers_for_ids()`.
- **`scaffold.py`** — emits per-surface files after the template copy, driven by selected `--agents` + installed MCPs; **skip-if-exists** so a re-scaffold never clobbers user-edited config.
- **`agent_guard_adapter.py`** — adds `cursor` + `antigravity` dialects (fail-open).
- **CLI** — `--agents` now accepts `cursor,antigravity,vscode`; `multi_agent` covers the GUI surfaces so the guard adapter ships. No per-surface template vars (would collide with the existing VS Code editor-settings var).

## What gets emitted

| Selection | Files |
|---|---|
| any MCP selected | root `.mcp.json` (Claude shareable scope) |
| `cursor` | `.cursor/hooks.json` (+ `.cursor/mcp.json` if MCPs) |
| `vscode` | `.vscode/mcp.json` (if MCPs) |
| `codex` | `.codex/config.toml` (if MCPs) — hooks/skills already via the codex overlay |
| `antigravity` (experimental) | `.agents/hooks.json` |

Skills are read natively by all of these — no per-surface skill rendering.

## Honesty / scope

Cursor/Antigravity **hooks are best-effort and fail-open**: the spike (#365) validated file paths + event names, but the exact blocking stdin/stdout I/O wasn't live-verified, so a wrong guess degrades safely and **git hooks + CI remain the enforcement boundary** (ADR-007). Antigravity is flagged experimental. **MCP emission is solid** (validated schemas).

## Tests

- `tests/unit/test_surfaces.py` — renderers, schema keys, experimental flags, selection→files.
- `tests/contracts/test_surface_emission.py` — scaffold-into-temp per surface, **drift guard** (emitted files must equal `planned_files`), skip-if-exists, adapter dialects.
- Full suite: **1034 passed, 3 skipped**; lint clean. README + `--agents` help updated.

Part of #359.
Closes #366


VytCepas:
**Review response (Codex — both P2s fixed):**
- **Stale generated config on re-scaffold:** Fixed. `surfaces.emit` no longer blunt-skips. Per file now: absent → write; present & identical to the fresh render → no-op; present & different → write a `.new` sibling (never clobber) and record it in the `conflicts` list so `upgrade` reports it. This neither clobbers a user edit nor silently leaves an old server list. (We can't distinguish a stale prior render from a user edit without storing the previous render, so any diff is surfaced for review — noted in the docstring.)
- **`.vscode/mcp.json` gitignored:** Fixed. The gitignore now uses the `.vscode/*` glob in both branches with `!.vscode/mcp.json`, so the generated VS Code MCP config is committable/shareable while personal editor files stay ignored. (A bare `.vscode/` ignore can't be negated per-file.)

Added tests: `.new`-on-change, no-op-when-unchanged, and VS Code MCP committable. 1036 tests green; lint clean.

PR #392, opened 2026-06-22
chore(PI-385): fix GUI-surface hook stdin parsing + deny shapes (Cursor, Antigravity)

### LINKED ISSUE #385: Live-verify Cursor + Antigravity GUI hook blocking contracts; promote from fail-open if confirmed
## Context
`src/project_init/surfaces.py` (PI-366 / ADR-017) already **emits** GUI-surface hook configs — `render_cursor_hooks()` (`.cursor/hooks.json`, `beforeShellExecution`/`beforeSubmitPrompt`) and `render_antigravity_hooks()` (`.agents/hooks.json`, `safety-gate`/`PreToolUse`). But per the module docstring (lines 10-13) the exact **stdin/stdout blocking decision contract** for these surfaces was *not live-verified* during the #365 spike, so both are shipped **best-effort and fail-open**, and Antigravity is flagged **experimental** (only `PreToolUse` confirmed).

## What to verify (live, against real installs)
- **Cursor** (≥1.7/2.4): does the shared guard (`agent_guard_adapter.py cursor`) actually *block* a denied shell command via `beforeShellExecution`? Confirm the expected exit-code / JSON decision schema and that a deny is honored (not just logged).
- **Antigravity** (v2.0): confirm `.agents/hooks.json` `safety-gate`/`PreToolUse` path is correct, the decision I/O contract, and whether more than `PreToolUse` is available. Verify the MCP location (currently treated as global `~/.gemini`, documented-not-emitted).

## Acceptance
- Document the verified contract for each in `surfaces.py` / the surface-matrix doc, raising confidence above the current MED/experimental.
- If a blocking contract is confirmed, make the adapter **fail-closed** for that surface (or explicitly record why it stays fail-open).
- If still unverifiable (Antigravity docs were JS-rendered/un-fetchable in the spike), keep experimental + fail-open and note that git/CI (ADR-007) remains the enforcement boundary.
- Add/extend a contract test asserting the emitted file matches the verified schema.

VytCepas:
Closes #385

## The bug
The shared guard adapter only parsed the Claude/Codex stdin shape (`tool_input.command`). Cursor sends the command **top-level** (`{"command"}`) and Antigravity sends it at `toolCall.args.CommandLine` — so for both GUI surfaces extraction returned empty and the hook fell open: **the Cursor and Antigravity command guards never actually blocked anything.**

## What this does
Contracts confirmed from vendor docs (Cursor `cursor.com/docs/hooks`; Antigravity via Google's migration guide), parsed/emitted, and pinned with simulation tests:
- **`_extract_command`** now handles Cursor's top-level `{"command"}` and Antigravity's `toolCall.args.CommandLine` (plus the existing Codex/Claude `tool_input` shape).
- **Cursor deny shape** corrected: `{"permission":"deny","user_message","agent_message"}` (was the wrong-case `userMessage`).
- **Dropped the dead `beforeSubmitPrompt` Cursor hook** — it carries no shell command and uses a different `{"continue":false}` deny shape the command guard can't produce, so it never did anything.
- **Antigravity** stdout `{"decision":"deny","reason"}` was already correct; now its stdin is parsed too.

## Decisions
- **Kept fail-open by design** (no Cursor `failClosed`). The ticket asked "promote from fail-open if confirmed" — the deliberate answer is *no*: per ADR-007 a hook crash must never wedge a user's shell; git/CI is the real boundary. Promoting to fail-closed would contradict the architecture.
- **Antigravity stays flagged `experimental`** — the *contract* is doc-confirmed and tested, but it isn't verified against a live `agy` binary.

## Tests
`TestAgentGuardAdapter` now feeds each surface's **real** stdin shape (Cursor top-level `command`; Antigravity `toolCall.args.CommandLine`; Codex `tool_input`) and adds a Cursor case proving `{permission:deny, user_message}`. `ruff` clean; **1043 passed / 3 skipped**.

## Honest limitation
I can't run Cursor/Antigravity binaries in this environment, so the contracts are **doc-confirmed + simulation-tested**, not live-binary-verified. If you want the live-binary check tracked, it's a small follow-up — but the never-blocked bug is fixed and the deny shapes match the published schemas.

### DIFF
--- a/docs/development/non-cli-surface-matrix.md
+++ b/docs/development/non-cli-surface-matrix.md
@@ -76,7 +76,7 @@ and `AGENTS.md`/`CLAUDE.md`/`GEMINI.md`.
 |---|---|---|
 | Claude Code VS Code ext | hooks, skills, CLAUDE.md (same `.claude/`) | **shared project MCP**: emit a root `.mcp.json` (`mcpServers`) when MCPs are configured — `.claude/` alone does not carry shareable MCP (bare `claude mcp add` writes a private `~/.claude.json` entry) |
 | Copilot agent mode | CLAUDE.md, `.claude/skills`, hooks (matcher-blind), AGENTS.md | `.vscode/mcp.json` (`servers` map) if MCPs are configured; document that matchers are advisory here |
-| Cursor | `.claude/skills` (skills), AGENTS.md | `.cursor/hooks.json` (translate `PreToolUse(Bash)`→`beforeShellExecution`, `UserPromptSubmit`→`beforeSubmitPrompt`); `.cursor/mcp.json` (≈copy of `mcpServers`) |
+| Cursor | `.claude/skills` (skills), AGENTS.md | `.cursor/hooks.json` (`PreToolUse(Bash)`→`beforeShellExecution`, deny via `{"permission":"deny","user_message"}`; PI-385); `.cursor/mcp.json` (≈copy of `mcpServers`) |
 | Codex (CLI/IDE) | `.codex/hooks.json` ✅, `.agents/skills` ✅, AGENTS.md ✅ | MCP → `.codex/config.toml` `[mcp_servers.*]` if MCPs are configured |
 | Antigravity | `.agents/skills` ✅, `.agents/hooks.json` ✅, `.agents/mcp_config.json` ✅, AGENTS.md ✅ (PI-386) | hook blocking-contract still experimental — live-verify in #385 |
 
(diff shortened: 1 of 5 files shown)


VytCepas:
**Review response (Copilot):** all 3 applied.
- **_extract_command crash on non-dict JSON:** Fixing — type-guarded `payload`/`tool_input`/`args`; non-dict stdin now returns `""` (fail-open) instead of raising. Added `test_malformed_payload_is_fail_open` (feeds `[]`) to pin it.
- **agent_message not asserted:** Fixing — the Cursor contract test now asserts both `user_message` and `agent_message`.
- **stale matrix Cursor row:** Fixing — the gap-table row no longer says `UserPromptSubmit→beforeSubmitPrompt` (that hook was dropped); it now documents the `beforeShellExecution` deny shape.

ruff clean; 1044 passed / 3 skipped.

VytCepas:
**Note:** Copilot's latest 3 comments are re-anchored duplicates of the original review — all three were already fixed in `7a548eb` (verified against current code): `_extract_command` type-guards `payload`/`tool_input`/`args` (lines 38/41/45; `test_malformed_payload_is_fail_open` feeding `[]` passes), the Cursor test asserts `agent_message` (line 238), and the matrix gap-table row no longer claims `UserPromptSubmit→beforeSubmitPrompt`. No further changes needed.

PR #415, opened 2026-06-22
feat(PI-410): governance overlay foundation + usage-track docs

VytCepas:
## Summary

First two increments of the **AI-governance epic (#276, ADR-018)**:

- **#410 — foundation:** the opt-in `--governance` overlay plumbing + layer skeleton + `governed` preset.
- **#411 — usage-track docs:** the pure-docs policy layer (no enforcement yet).

Also lands the **Codex-APPROVED epic plan** (`PLAN-276-governance.md` + review log) so the converged 4-round plan is preserved in history.

## #410 — foundation

- **`--governance` flag wired end-to-end**, mirroring `--multi-model` at every site: argparse, `ScaffoldInputs`, interactive chooser, `cli_overlays`, both ctors, `_build_variables`, `overlay_layers(governance=)`, upgrade off-default + re-render, capabilities inventory, helpers, README.
- **Dual enablement**: CLI flag **or** the `governed` preset's `[vars] governance = true`; the flag wins. Resolution mirrored at the `overlay_layers` call site **and** in the recorded variable, so the appended layer and `config.yaml` can never disagree (Codex r1 #14 / r2 #3).
- `templates/governance/` layer skeleton + `governed` preset (`extends obsidian-only`).

## #411 — usage-track docs

Adopt-and-customise policy templates that **adopt** NIST AI RMF / ISO 42001 / EU AI Act / OWASP, under `.claude/governance/`:

- `AI_USAGE_POLICY.md` — 1-page umbrella AUP.
- `approved-tools.md` — allow/deny policy (deny-by-default); links `CAPABILITIES.md` as supporting inventory only (Codex #6).
- `data-handling.md` — data-class matrix + gitleaks/`--no-egress`/`prod_guard` backstops.
- `ai-code-provenance.md` — attribution, licence contamination, review bar.
- `NIST_RMF_CROSSWALK.md` — Govern/Map/Measure/Manage + ISO/EU/OWASP pointers.

## Tests

- `tests/contracts/test_governance_overlay.py`: overlay on/off, preset merge, flag/preset/off resolution, **upgrade round-trip** (layer survives re-render from the recorded variable alone), and all five usage-track docs scaffold with their load-bearing content.
- Fixed `test_presets_have_required_keys` to resolve `extends`.
- Full suite: **1070 passed, 3 skipped**; `ruff check` clean.

## Not in scope (follow-ups)

- #412 — system card + two-file AIBOM + presence-triggered CI gate
[...]

Issue #413, opened 2026-06-22
Governance: ADR-018 (three-bucket split, system-card + gate conventions, why Phase 4 rejected)


VytCepas:
Part of #276 (AI governance epic). **Phase 5 — record the decisions.**

Add `docs/adr/adr-018-ai-governance-layer.md` (next free number; confirm at write time) recording:
- The **three-bucket split** (Bucket 1 built here; Bucket 2 cloud-IaC fork and Bucket 3 org registry are boundaries only).
- **System card, not model card** + the broadened governance-manifest fields.
- The **split `classification` + `allowed`** convention with `prohibited` = hard fail (`exception` is record-only).
- **Presence-triggered, multi-field, CI-first** gate; flat-scalar parsing (no PyYAML).
- **Derive AIBOM from `mcps.py` + a CCR extractor**, split into generated vs user-owned (preserve-glob) files; detected-vs-declared labelling.
- **Why Phase 4 (risk_tier-as-orchestration) was rejected** — overlay composition is fixed at scaffold/upgrade time from recorded variables, so a later-edited card can't drive it; it also depended on an unbuilt observability overlay. Any future runtime use of classification reads `.claude/config.yaml`, not markdown.

## Acceptance
- ADR-018 added and listed in the ADR index; `just docs` builds clean.

Issue #411, opened 2026-06-22
Governance: usage-track docs layer (AUP, approved-tools, data-handling, code-provenance, NIST crosswalk)


VytCepas:
Part of #276 (AI governance epic). **Phase 1 (usage track) — the MVP. Depends on the `--governance` flag + layer skeleton sub-issue.**

Ship the usage-governance docs into `templates/governance/dot_claude/governance/` — pure templated docs, no enforcement. Covers how staff *use* AI tools (the shadow-AI / data-leakage risk), the cheapest highest-coverage governance content.

## Scope (files)
- `AI_USAGE_POLICY.md` — 1-page AUP, NIST/Responsible-AI-Institute-aligned, committed + versioned.
- `approved-tools.md` — explicit **allow/deny policy** (sanctioned models, endpoints, allowed data classes, use cases). Links `CAPABILITIES.md` only as supporting inventory — it is NOT the policy (Codex r1 #6).
- `data-handling.md` — what data may/may not go to AI tools (PII, secrets, customer data, NDA source); references the already-enforced gitleaks + `--no-egress` controls.
- `ai-code-provenance.md` — AI-authored-code policy: attribution, license contamination, review requirements.
- `NIST_RMF_CROSSWALK.md` — maps each artifact to Govern/Map/Measure/Manage (adopted reference, dated; treat as reference not authored truth).
- Focused pytest scaffolding the governance layer into a temp dir.

## Acceptance
- `--governance` scaffolds all five docs under `.claude/governance/`.
- Focused pytest asserts presence + that `approved-tools.md` is allow/deny structured.
- `just lint && just test` green.

Issue #410, opened 2026-06-22
Governance: --governance flag plumbing + governed preset + layer skeleton


VytCepas:
Part of #276 (AI governance epic). **Foundation sub-issue — Phase 2 of the approved plan; the docs layer (sub) and gate (sub) depend on this.**

Wire an opt-in `--governance` overlay through the scaffolder and add a `governed` preset, plus the empty `templates/governance/` layer skeleton the content/enforcement sub-issues fill in.

## Scope
- New `templates/governance/` layer composed via `overlay_layers()`.
- `--governance` flag mirroring `--multi-model` at **every** site: argparse; `ScaffoldInputs` field; interactive chooser (rich panel, default False); `cli_overlays` tuple + unpack; resolve; both `ScaffoldInputs(...)` ctors; **`_build_variables()` render var** `"governance": "true" if inputs.governance else ""`; scaffold-time `overlay_layers(... governance=...)`.
- `scaffold.py`: `overlay_layers()` gains `governance: bool = False` → appends `"governance"`.
- `upgrade.py`: `"governance": ""` in `_overlay_off_defaults()`; `governance=bool(variables.get("governance"))` in the `_render_staging` `overlay_layers(...)` call.
- `capabilities.py`: inventory entry; `tests/helpers.py make_variables()` += `"governance": ""`; `README.md` documents the flag.
- **`governed` preset** in `templates/presets/`: because `_build_variables()` only threads `memory_stack` from `preset["vars"]` (`__main__.py:1023`), teach it to resolve `governance` from `inputs.governance` **or** `preset["vars"].get("governance")`, **CLI flag taking precedence** (Codex r2 #3).

## Acceptance
- `overlay_layers("claude", no_plugin=False, governance=True) == ["governance"]`; off ⇒ omitted; composes with agents/fallback/multi_model.
- `upgrade --apply` round-trip re-derives the layer from the recorded variable; pre-feature record re-renders faithfully via `_overlay_off_defaults()`; no spurious `.new`.
- `governed` preset records `governance: true` and an **upgrade round-trip test proves the layer survives re-render**.
- `just lint && just test` green.

PR #416, opened 2026-06-23
feat(PI-412): governance product track — system card + AIBOM + CI gate

### LINKED ISSUE #412: Governance: system card + derived AIBOM + presence-triggered enforcement gate
Part of #276 (AI governance epic). **Phase 3 (product track) — the load-bearing differentiator: a failing CI check, not a PDF. Depends on the `--governance` flag + layer skeleton sub-issue.**

Ship the system card + derived AIBOM + presence-triggered enforcement gate.

### System card (governance manifest)
- `SYSTEM_CARD.example.md` + `SYSTEM_CARD.template.md` under `examples/` — an **example/template, NOT a live gated card**, so a fresh scaffold passes (Codex r1 #7).
- **Flat-scalar frontmatter only** (no PyYAML, no nesting — Codex r1 #9): `owner`, `system_name`, `role` (`provider|deployer`), `use_case`, `affected_users`, `data_classes`, `human_oversight` (`yes|no|n-a`), `logging` (`yes|no|n-a`), `last_reviewed` (ISO date).
- `classification` (`prohibited|high|limited|minimal`) + `allowed` (`true|false`): **`prohibited` always requires `allowed:false`** — `prohibited`+`allowed:true` is always a hard gate failure; `exception:` is record-only evidence, never run permission (Codex r1 #2, r2 #4).
- `models_declared` = a **repo-root-relative path reference** to `ai-declarations.md` (must resolve to/under `.claude/governance/`; absolute/traversal rejected) — one source of truth, not an inline list (Codex r2 #5, r3 #3).
- Prose sections (purpose, intended/prohibited use, limitations, oversight) + a dated EU-AI-Act/NIST mapping checklist (prose, not logic).

### AIBOM — two files, never mixed (Codex r1 #4/#5, r2 #2, r3 #1)
- `ai-bom.generated.md` — regenerated/overwritten each scaffold+upgrade. MCP inventory from `project_init.mcps.servers_for_ids()` over `variables["installed_mcps"]` (NOT surfaces.py); model inventory via a **fixture-tested CCR extractor** reading `multi_model/config.json` `Providers`/`Router` only when present, labelled **"detected CCR routes"** (empty + note when absent).
- `ai-declarations.md` — **user-owned, seeded once** (write-if-absent like `scaffold.py:325`) + registered in **preserve globs** (`read_preserve_globs`/`_is_preserved`, `upgrade.py:362-366`) so upgrade never overwrites it. **Merge with any existing `preserve:` array, don't replace** (Codex r4). Holds hand-declared models/APIs CCR can't see.

### Gate — presence-triggered, multi-field, CI-first
- `governance_gate.sh` — clone of `pre_commit_gate.sh` shape, Python via `_py.sh`. For each real `SYSTEM_CARD.md` (glob, excluding example/template), fail if: required field missing/placeholder/empty; `classification`/`allowed`/`role`/oversight out of allowed sets; illegal `prohibited`+`allowed:true`; `models_declared` points at a missing/placeholder/escaping `ai-declarations.md`; or `last_reviewed` past the staleness window.
- **Staleness anchored** (Codex r2 #6, r3 #4): hardcoded default (e.g. 180d) overridable only via flat scalar `staleness_days` in `.claude/governance/config.json` — ship `config.example.json` only; real `config.json` is user-owned (preserve glob); default applies when absent.
- CI `governance` job is the enforcement boundary (git+CI principle). `templates/fallback/` local pre-commit hook is best-effort; a static no-op-unless-card plugin hook is an optional follow-up — v1 is CI-first (Codex r1 #10).

## Acceptance
- Gate passes on a valid card; fails on each violation class; no-ops with zero real cards; rejects absolute/traversal `models_declared` (tested).
- CCR extractor unit-tested against a fixture config; AIBOM split files behave correctly across upgrade (generated overwritten, declarations preserved).
- Tests mirror `tests/integration/test_hooks_and_safety.py`. `just lint && just test` green.

VytCepas:
## Summary

Phase 3 of the AI-governance epic (**#276, ADR-018**) — the **product track**, layering enforcement on top of the policy docs from #411. Builds on the merged foundation (#410/#411, PR #415).

## What's in it

**System card** (`.claude/governance/examples/`)
- `SYSTEM_CARD.example.md` (a valid reference) + `SYSTEM_CARD.template.md` (copy-me) carrying a flat-scalar **governance manifest**: `classification` + `allowed` split (`prohibited` always requires `allowed:false`), `role`/`human_oversight`/`logging` enums, `models_declared` as a path reference. Examples are excluded from the gate, so a fresh project passes.
- `config.example.json` — staleness override shipped as an *example only*; the real `config.json` is user-created.

**AIBOM — two files, never mixed**
- `src/project_init/governance.py`: `emit()` writes `ai-bom.generated.md` (installed MCPs via `mcps.servers_for_ids` + a **fixture-tested CCR extractor** labelled "detected CCR routes"), regenerated each scaffold/upgrade like `CAPABILITIES.md`.
- `ai-declarations.md`: user-owned, **seeded once and preserved** via `config.yaml` preserve globs rendered under `{{#if governance}}` — no runtime merge (config.yaml is itself preserved on upgrade, sidestepping the merge/malformed-parse concern from Codex r4).

**Gate** (`.claude/scripts/governance_gate.{sh,py}`, stdlib via `_py.sh`)
- Presence-triggered, multi-field: required fields, allowed-value sets, the `prohibited`+`allowed:true` illegal combo, `models_declared` **path containment** (rejects absolute / `..` traversal) + completeness, `last_reviewed` staleness (180d default, `config.json` override). **No real card ⇒ pass.**
- `ci.yml.tmpl` gains a `{{#if governance}}` **governance job** — the enforcement boundary (matches the repo's git+CI principle).

**Refactor:** extracted `scaffold()`'s post-copy generation into `_emit_generated_files` to stay under the C901 complexity limit.

## Tests

`tests/contracts/test_governance_product.py` (18 cases): CCR extractor (fixture + absent-file), AIBOM on/off + CCR detection, declarations seed/preserve round-trip + preserve-glob registration, and the gate across pass + **every violation class** (missing field, prohibited combo, out-of-range role, staleness + config override, unfilled declarations, path traversal/absolute) as real subprocesses.

Full suite: **1097 passed, 3 skipped**; `ruff check` clean. Verified end-to-end by scaffolding a `governed` project, exercising the gate (pass/fail), and an `upgrade` round-trip (declarations preserved, AIBOM regenerated, CI job renders only when on).

## Follow-up

- #413 — ADR-018 write-up (records this design; observability moved to ADR-019).

Part of #276
Closes #412


VytCepas:
**Review response (Codex + Copilot, cycle 1):**

- **[Codex P2] config.yaml preserve globs don't reach projects that adopt governance later** — valid; the config-render approach has exactly that gap (config.yaml isn't re-rendered once it carries a record). Switched to Codex's preferred fix: the three user-owned governance files (`SYSTEM_CARD.md`, `ai-declarations.md`, `config.json`) are now **intrinsically preserved** via a `_GOVERNANCE_USER_FILES` set checked in both `_should_preserve` (scaffold) and `_is_preserved` (upgrade), and the `{{#if governance}}` config.yaml render was **removed**. The generated `ai-bom.generated.md` is deliberately excluded so it still refreshes. Tests updated + an explicit upgrade-preservation test added.
- **[Copilot] `find_cards` excludes any `examples` segment** — fixed to exclude only the top-level `.claude/governance/examples/` (`parts[0] == 'examples'`); a card under a deeper `examples/` is still gated. Added a test.
- **[Copilot] future `last_reviewed` bypasses staleness** — fixed: negative age is now rejected as 'in the future'. Added a test.
- **[Copilot] `_emit_generated_files` `conflicts` type** — annotated `list[...] | None` to match `surfaces.emit`.
- **[Copilot] test write_text default encoding** — all card read/writes in the test now pass `encoding='utf-8'`.

Full suite: 1100 passed, 3 skipped; ruff clean.

Issue #407, opened 2026-06-22
Observability: docs (using/upgrading guides), #269 tie-in, ADR-019


VytCepas:
Part of #269 (Track A — observability overlay; epic #403 was collapsed into #269).

> **ADR numbering:** ADR-018 is taken by governance epic #276 (issue #413, lands first). Observability uses **ADR-019**.

### Build
- `dot_claude/docs/guides/using-observability.md` — Claude-Code-only scope, local-only/no-egress, snapshot model, the 4 buckets, approximate-vs-exact caveat.
- `dot_claude/docs/guides/upgrading-observability.md` — OPTIONAL OTEL→Grafana/Phoenix upgrade path. **Doc only, no code.**
- `docs/development/measurement-methodology.md` — note the future `tools/benchmark/` #269 harness reuses this analyzer's parsing over the same local sources (repo harness may use `rich`; scaffolded analyzer stays stdlib-only).
- `docs/adr/adr-019-local-observability-overlay.md` — record the file-based / no-Docker / no-OTEL decision + documented upgrade path.

### Acceptance
- Guides scaffold under the overlay; docs build clean; ADR-019 added to the index.

VytCepas:
## Fresh-pass Codex plan review — finding for #407

**P2 — docs must be explicit about scope: Claude Code only; Amp/Junie have no hooks to log.**
Amp and Junie are now valid `--agents` surfaces (`src/project_init/__main__.py:160`) but are **MCP-only / no-hook** (`src/project_init/surfaces.py:168` `"hooks_file": None`, and the Junie entry). The hook self-log (#406) therefore cannot capture anything for them.
*Amend:* the using/upgrading guides + ADR-018 should state plainly that observability covers **Claude Code transcript + Claude hook self-log only**; other surfaces (Amp/Junie/Cursor/Antigravity/Codex) are out of scope for usage capture, by design.

Issue #405, opened 2026-06-22
Observability: stdlib analyzer (usage_report.py) + observability.sh report


VytCepas:
Part of #269 (Track A — observability overlay; epic #403 was collapsed into #269).

The new `templates/observability/` overlay layer (mirrors `multi_model`; **static files only**, no `.tmpl`).

### Build
- `dot_claude/observability/usage_report.py` — **stdlib-only** analyzer (no third-party imports; runs via `_py.sh`). **Aggregate-only by construction**: extracts only names/counts/numbers — never prompt text, tool-input bodies, command strings, or file contents.
  - **Transcript discovery** (order): `--transcript <path>` / `--session-id <id>` → derive Claude project slug from project dir (`/`→`-`, leading `-`, `.`→`-`) reading `~/.claude/projects/<slug>/*.jsonl` → fallback glob `~/.claude/projects/*/*.jsonl` filtered by the `cwd` field in transcript entries. Never silently empty — clear "not found, pass --transcript" error.
  - 4 buckets: **Adoption** (per-skill/tool/sub-agent/hook counts), **Cost** (USD via embedded static price table + tokens, cache-read ratio; per-skill cost = time-window heuristic, labelled approximate), **Productivity** (LoC approx from Edit/Write diffs; commits from **local git only**; PRs from local merge-commit heuristics or omitted — **no `gh`/network**), **Reliability** (success/error from `tool_result.is_error`; accept/reject + exact active-time noted OTEL-only/absent).
  - Output: text summary to stdout + **self-contained** `dashboard.html` (inline SVG/CSS, **no CDN/JS**, no external URL).
- `dot_claude/scripts/observability.sh` — `+x`; `report [--open] [--transcript …] [--session-id …]`; resolves Python via `_py.sh`; `--open` best-effort (xdg-open/open/explorer), fail-open.
- `dot_claude/observability/.keep` — the **activation marker** (`.claude/observability/` presence flips the hook self-log on; see sibling issue).
- `dot_claude/observability/README.md`.

### Acceptance
- Imports no third-party module; `observability.sh` executable bit set.
- Fixture transcript + usage.jsonl → all 4 buckets; known slug forms resolve.
- `dashboard.html` contains **no external URL** and **no raw transcript content** (assert fixture prompt/command/path text does not leak).
- No networked command invoked (no `gh`).

VytCepas:
## Fresh-pass Codex plan review — findings for #405

[...]

Issue #269, opened 2026-06-18
Metrics & evaluation: agent observability overlay + scaffold cost–benefit benchmark


VytCepas:
## Summary

Single epic for **all metrics & evaluation work**. Two previously-separate epics — #269 (repo-internal "is the scaffold worth it?" benchmark) and #403 (shipped `--observability` overlay) — are merged here, because they answer the same family of questions over the **same local data** and were already coupled by an explicit "#403 → #269 tie-in".

> **Merge note:** #403 has been collapsed into this epic (closed as superseded). Its four sub-issues (#404–#407) now hang directly off #269. Its locked plan + Codex adversarial-review transcript (`PLAN.md` / `PLAN-REVIEW-LOG.md`) still apply to Track A and should be attached to the first Track A PR.

The headline question is **not** "how many tokens?" but **"is the scaffold worth it, and is anyone using it?"** — pairing every cost delta with the quality/adoption delta it bought.

**Hard constraint (CLAUDE.md / ADR-007):** the scaffolder stays deterministic and **never calls an LLM**. Repo-internal benchmark tooling lives in `tools/` or `tests/benchmarks/`; the shipped overlay analyzer is stdlib-only and runs on demand inside scaffolded projects. Neither is part of the scaffold runtime.

## Why these two merge

Both consume the **same sources**: Claude Code's per-session transcript JSONL (`input_tokens` / `output_tokens` / `cache_creation_input_tokens` / `cache_read_input_tokens` + timestamps) plus a guarded hook self-log, and derive $ from a static model-keyed price table. Both count skill/tool/sub-agent/hook usage. They differ only in **audience and output**:

| | **Track A — Observability overlay** (was #403) | **Track B — Cost–benefit benchmark** (was #269 core) |
|---|---|---|
| Audience | End-users of a scaffolded project | Maintainer / fork owner of this repo |
| Question | "Are the installed skills/hooks/agents actually used? cost/reliability?" | "Does the preset earn its overhead vs a bare project?" |
| Form | Shipped `templates/observability/` overlay; text + `dashboard.html` | `tools/`/`tests/benchmarks/` runner + `rich` Pareto verdict |
| Egress | Zero (transcript + local git only) | Local only |

## Shared parsing core (build once)

[...]

PR #421, opened 2026-06-23
feat(PI-406): guarded hook self-log (_usage_log.sh + prod_guard.py)

### LINKED ISSUE #406: Observability: guarded hook self-log (_usage_log.sh stdin-safe + prod_guard.py + sync)
Part of #269 (Track A — observability overlay; epic #403 was collapsed into #269).

The one signal neither transcript nor OTEL records cleanly: hook firings. **Shipped-always-dormant** — existing always-on hooks call a guarded helper that no-ops unless the overlay's `.claude/observability/` marker exists (the plugin `hooks.json` is static / non-gateable, so no new hook + no overlay-only wiring).

### Build
- `templates/fallback/dot_claude/hooks/_usage_log.sh` — `usage_log <hook> <event> [cwd]`:
  - (a) **NEVER reads stdin** (would consume the hook JSON before `dag_workflow.py` etc.). Resolve project root from `$CLAUDE_PROJECT_DIR` → `git rev-parse --show-toplevel` → caller-passed `[cwd]` arg only.
  - (b) early-return unless `$root/.claude/observability/` exists.
  - (c) else append one JSON line `{ts,hook,event,project[,session]}` to `$root/.claude/logs/usage.jsonl` (`mkdir -p`). `session` is **optional** — emit only if a known env (e.g. `$CLAUDE_SESSION_ID`) exists; analyzer joins by timestamp+project otherwise.
- The five shell hooks (`session_setup`, `pre_commit_gate`, `github_command_guard`, `post_edit_lint`, `workflow_state_reminder`) source the helper + call `usage_log` once (passing cwd if already cached).
- `prod_guard.py` (Python, `base/`, run via `_py.sh`): parse stdin payload **once in `main()`**, pass the dict to **both** a new inline stdlib `usage_log()` and `evaluate()` — no second stdin read.
- **Sync:** add `_usage_log.sh` to `tools/sync_plugin.py` fallback-hook set; `uv run python tools/sync_plugin.py`; assert no drift. Plugin `hooks.json` unchanged.

### Acceptance
- Self-log fires in **default plugin mode** AND `--no-plugin` mode when marker present; silent when absent.
- **Stdin passthrough preserved**: payload still reaches `dag_workflow.py`; `prod_guard.py` still **blocks a destructive command with observability ON** (regression test).
- `sync_plugin.py` no drift.

VytCepas:
Closes #406. Third increment of epic #269 Track A (observability overlay); the writer side of the `usage.jsonl` that #405's analyzer already reads.

## What

Captures the one signal neither the transcript nor OTEL records cleanly — **hook firings** — **shipped-always-dormant**: the helper and wiring ship in every scaffold but no-op unless the overlay marker `.claude/observability/` exists. The plugin `hooks.json` is static / non-gateable, so the gate lives in the hook body rather than in separate wiring.

- **`_usage_log.sh`** (fallback hooks) — defines `usage_log <hook> <event> [cwd]`:
  - **NEVER reads stdin** — the hook payload JSON belongs to the real hook body (`dag_workflow.py`, `pre_commit_gate`, …); consuming it would starve them. Inputs come from args + env only; root resolves `$CLAUDE_PROJECT_DIR` → `git rev-parse` → `[cwd]` → `$PWD`.
  - **marker-gated** (dormant unless `.claude/observability/` exists), **fail-open**, appends one JSON line `{ts,hook,event,project[,session]}`. `session` only when `$CLAUDE_SESSION_ID` is set.
- **The five always-on shell hooks** (`session_setup`, `pre_commit_gate`, `github_command_guard`, `post_edit_lint`, `workflow_state_reminder`) source it and call `usage_log` once — before any stdin read / `exec`, with `</dev/null` so the payload reaches the real body intact.
- **`prod_guard.py`** — new `usage_log()` folds off the **already-parsed** payload (no second stdin read), marker-gated, fail-open; called in `main()` before `evaluate()`. Uses portable `time.gmtime()` (no `datetime.UTC`, which is 3.11+ — scaffolded projects may run older Python).
- **Sync:** `sync_plugin.py` auto-discovers the new helper; plugin copies re-synced. No drift.

## Acceptance (all green)
- Self-log fires in **default plugin mode** AND **`--no-plugin`** mode when the marker is present; **silent when absent** (verified by scaffolding both ways).
- **Stdin passthrough preserved** — `usage_log` reads `</dev/null`; a subprocess test confirms stdin survives the call. `prod_guard.py` **still blocks** `terraform destroy` with observability ON (regression test).
- `sync_plugin.py` reports no drift; the existing `test_hook_scripts_in_sync` covers the new helper.
- `just lint && just test` green (1151 passed, 3 skipped); smoke-tested end-to-end.

## Note for reviewers
Writes to **`.claude/observability/usage.jsonl`** (next to the marker, where #405's analyzer + the layer `.gitignore` already point) — **not** the issue's pre-#405 `.claude/logs/` path. The issue text was written before #405 fixed the placement; this keeps writer and reader consistent.

Follow-up: #407 (using/upgrading docs + ADR-019) closes out Track A.

### DIFF
--- a/plugins/project-init-workflow/hooks/_usage_log.sh
+++ b/plugins/project-init-workflow/hooks/_usage_log.sh
@@ -20,11 +20,19 @@
 # is emitted only when $CLAUDE_SESSION_ID is set; the analyzer otherwise joins
 # by timestamp + project. Fully fail-open: any error is swallowed.
 
-# _usage_log_json_escape <string> — minimal JSON string escaping (\\ and ").
+# _usage_log_json_escape <string> — JSON string escaping. Backslash and quote
+# first, then the control chars that would otherwise split the line or produce
+# invalid JSONL (a tab/newline in a path or $CLAUDE_SESSION_ID). $'...' is
+# ANSI-C quoting, available on bash 3.2.
 _usage_log_json_escape() {
   local s=$1
   s=${s//\\/\\\\}
   s=${s//\"/\\\"}
+  s=${s//$'\n'/\\n}
+  s=${s//$'\r'/\\r}
+  s=${s//$'\t'/\\t}
+  s=${s//$'\b'/\\b}
+  s=${s//$'\f'/\\f}
   printf '%s' "$s"
 }
 
--- a/plugins/project-init-workflow/hooks/_usage_log.sh
+++ b/plugins/project-init-workflow/hooks/_usage_log.sh
@@ -1,0 +1,66 @@
+#!/usr/bin/env bash
+# _usage_log.sh — guarded hook self-log for the observability overlay (ADR-019,
+# #406). Sourced by the always-on shell hooks; defines usage_log().
+#
+# SHIPPED-ALWAYS-DORMANT: every scaffold carries this helper, but it no-ops
+# unless the observability overlay's marker directory (.claude/observability/)
+# exists — so it costs nothing until a project opts in by scaffolding the
+# overlay. The plugin hooks.json is static / non-gateable, hence this in-hook
+# guard rather than separate wiring.
+#
+# CRITICAL — never reads stdin. The hook payload JSON on stdin belongs to the
+# real hook body (dag_workflow.py, pre_commit_gate, …); consuming it here would
+# starve them. Inputs come from args + env only:
+#   usage_log <hook> <event> [cwd]
+# Project root is resolved from $CLAUDE_PROJECT_DIR, else `git rev-parse`, else
+# the optional [cwd] arg, else $PWD — git/network are never required.
+#
+# Appends one JSON line {ts,hook,event,project[,session]} to
+# .claude/observability/usage.jsonl (the file usage_report.py reads). Session id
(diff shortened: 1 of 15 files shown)


VytCepas:
**Review response** (Copilot — all three applied):

- **[_usage_log.sh:29 ×2] JSON escaping missed control chars**: Fixing — `_usage_log_json_escape` now also escapes `\n \r \t \b \f` (after backslash/quote), so a tab/newline in a path or `$CLAUDE_SESSION_ID` can never split the JSONL line or produce invalid JSON. Fixed in the fallback source + re-synced to the plugin copy (both identical). Added a regression test: `CLAUDE_SESSION_ID="ab\tcd\nef"` → exactly one physical line, valid JSON, value round-trips.
- **[test:37] test helper not shell-quoted**: Fixing — `_call_helper` now `shlex.quote`s the helper path and every arg before embedding them in the bash snippet, so a tmp path with spaces/metachars can't mis-execute.

`just lint && just test` green (1152 passed, 3 skipped); plugin/source confirmed identical (no drift).

Issue #275, opened 2026-06-18
Cost–benefit presentation layer: terminal report with a Pareto-aware &quot;is it worth it?&quot; verdict


VytCepas:
## Summary

The **headline deliverable**: a readable terminal report that turns the captured dimensions into a **cost–benefit verdict** — not raw numbers. It answers *"is the scaffold worth it?"* by pairing every cost delta with the quality delta it bought.

Display (rich tables/panels; `rich` is already a dependency):
- **Bare vs scaffolded, side by side, with deltas** — token/cost, latency, accuracy, efficiency. Each cost increase shown next to what it bought (`+18% tokens → −40% wrong turns, +25% first-try test pass`).
- **Plain-language verdict line** per target/preset — "*costs X more, buys Y*" — plus a **Pareto flag**: efficient vs **dominated** (more cost, no better outcome).
- **Per-artifact attribution** of fixed overhead — which always-loaded scaffold files (CLAUDE.md, hooks, reminder, INDEX/skills) cost the most context, so they can be trimmed.
- **Diminishing-returns view** across presets/overlays (base → obsidian → graphify): is the heavier preset still on the cost–accuracy frontier, or past the knee?
- Optional: also emit a written report (markdown) for docs; terminal is the minimum.

Grounding: agent-eval practice scores **both** quality and cost on a Pareto frontier and surfaces diminishing returns — measuring either axis alone is misleading. This layer must never be confused with the scaffold runtime; it reads harness artifacts only.

## Metadata

- Type: feature
- Priority: medium
- Size: M
- Area: tools/ (rich-based report)
- Agent ready: No
- Confidence: medium

## References

- Parent epic: #269
- Cost–accuracy Pareto framing: https://cobusgreyling.substack.com/p/the-pareto-frontier-for-ai-agents
- Diminishing returns / pricing mistakes: https://cognaptus.com/blog/2025-07-08-beyond-the-pareto-frontier-pricing-llm-mistakes-in-the-real-world/

## Dependencies

- All capture children (token/cost, latency, accuracy, efficiency) — consumes their records

## Acceptance criteria

- [ ] Terminal report shows bare vs scaffolded with deltas across all captured dimensions.
- [ ] Every cost delta is paired with the quality/efficiency delta it bought (not shown in isolation).
- [ ] A per-target verdict line + Pareto flag (efficient/dominated) is rendered.
- [ ] Fixed-overhead per-artifact attribution is shown.
[...]

Issue #271, opened 2026-06-18
Collection harness: run representative tasks in bare vs scaffolded targets and capture usage


VytCepas:
## Summary

Build the **runner** every other dimension feeds. It executes the methodology's representative tasks against a **bare** target and one or more **scaffolded** targets, then captures the raw signals (token usage, timestamps, tool calls) into a structured artifact (e.g. JSON/JSONL) the presentation layer can consume.

Scope:
- Set up bare and scaffolded target dirs (scaffold via `project-init` into temp dirs, matching the existing template-test pattern).
- Drive the agent on each task (or parse an existing run), then extract usage from the chosen data source — Claude Code transcripts are JSONL with per-message `usage` (`input_tokens`, `output_tokens`, `cache_creation_input_tokens`, `cache_read_input_tokens`) plus timestamps for wall-clock.
- Emit one normalized record per (task, target, run) so dimensions and the report read a stable schema, not raw transcripts.
- Lives in `tools/` or `tests/benchmarks/` — **never** invoked from the scaffold runtime (CLAUDE.md: no LLM calls from the scaffolder).

## Metadata

- Type: feature
- Priority: medium
- Size: L
- Area: tools/ or tests/benchmarks
- Agent ready: No
- Confidence: medium

## References

- Parent epic: #269
- Depends on the methodology issue (task set + data source)

## Dependencies

- Methodology + metric taxonomy issue (blocks this)

## Acceptance criteria

- [ ] Harness scaffolds/sets up bare and scaffolded targets for the defined task set.
- [ ] Per-run usage + timing + tool-call signals are extracted into a normalized record schema.
- [ ] Output is a stable artifact the presentation layer can read without re-parsing transcripts.
- [ ] No scaffold runtime change; tooling isolated under tools/ or tests/benchmarks.

## Definition of Ready

- [ ] Methodology issue merged (task set + data source pinned).

## Definition of Done

- [ ] Harness produces normalized records for at least one task across bare + scaffolded; documented how to run it.

VytCepas:
**Reuse #403's analyzer.** Per the unification note on #269: extract transcript JSONL via #403's `usage_report.py` (#405) — import/wrap it, don't fork a second parser. Depends on #405 landing first.

PR #425, opened 2026-06-23
feat(PI-273): accuracy/task-success scoring (scoring.py)

### LINKED ISSUE #273: Accuracy / task-success scoring: define and capture correctness signals per task
## Summary

Capture the **benefit** axis — did the agent actually succeed? Without this, the cost numbers are meaningless (cheap-but-wrong is not a win).

Per the methodology's per-task success definition, record a deterministic correctness signal for each run:
- **Task success** — did the deterministic check pass (tests green / output matches expected / required files produced)?
- **First-try vs retries** — succeeded on first attempt, or only after N corrections?
- **Review/rework cycles** — how many correction rounds were needed (ties to the lifecycle review protocol).

Keep scoring deterministic where possible (test exit codes, file/diff checks). If any judged signal is needed, it is computed in tooling, not in the scaffold runtime.

## Metadata

- Type: feature
- Priority: medium
- Size: M
- Area: tools/ or tests/benchmarks
- Agent ready: No
- Confidence: medium

## References

- Parent epic: #269
- Methodology issue (success definition per task)

## Dependencies

- Methodology issue (accuracy definition); collection harness (records)

## Acceptance criteria

- [ ] Each run records a deterministic task-success boolean per its defined check.
- [ ] First-try vs retried, and rework-cycle count, are captured.
- [ ] Scoring is reproducible from the run artifact (no manual judgement required for the core signal).

## Definition of Ready

- [ ] Accuracy/success signal defined in the methodology issue.

## Definition of Done

- [ ] Success + retry/rework fields present per record; test asserts scoring on a known pass/fail fixture.

VytCepas:
Closes #273. The **benefit axis** for epic #269 Track B — without it the cost numbers are meaningless (cheap-but-wrong is not a win). Fills `success`/`first_try`/`rework_cycles` on the harness records via the `[check]` specs authored in #271. No LLM judge (CLAUDE.md / ADR-001).

## What

- **`scoring.py`** — `score(task, *, target_dir, transcript_path, agent_output) -> Score`:
  - **`success`** — deterministic per the task's `[check]`: `pytest` exits as expected **and** required files exist (`pytest` — `feat`/`fix`); the agent's final text matches a pattern (`regex` — `qa`); or `None` when no check applies (`none` — the `noop` probe).
  - **`rework_cycles`** — count of error `tool_result` blocks in the transcript: a deterministic, artifact-reproducible **proxy** for correction rounds.
  - **`first_try`** — succeeded with zero rework.
- **`transcript.py`** — `parse` now also counts error `tool_result`s (`tool_errors`) by folding user tool-result blocks.
- **harness** — `run_task` returns the agent's final text (for the regex check); `run` fills all three accuracy fields on every record; `record-from` fills `rework_cycles` from the transcript (no target → `success`/`first_try` stay null; use `run` for the full score).

## Acceptance (all green)
- Each run records a **deterministic task-success boolean** per its defined check.
- **First-try vs retried + rework-cycle count** captured.
- **Scoring is reproducible from the run artifact** — no manual judgement for the core signal.
- **Test asserts scoring on a known pass/fail fixture**: pytest passes/fails/missing-file, regex match/no-match, none→None, rework count, first_try logic. Verified the `fix` task end-to-end — the score flips **False→True** when `calc.py` is repaired.
- `just lint && just test` green (1200 passed, 3 skipped).

## Last on Track B
The record now carries all three axes (cost, latency, accuracy). **#275** renders the Pareto-aware `rich` "is it worth it?" verdict — bare vs scaffolded, each cost delta paired with the quality it bought — consuming `cost_usd` + `latency.summarize` + these accuracy fields.

### DIFF
--- a/tests/contracts/test_benchmark_scoring.py
+++ b/tests/contracts/test_benchmark_scoring.py
@@ -1,0 +1,111 @@
+"""#273: accuracy / task-success scoring against the authored [check] specs.
+
+Deterministic — pytest exit codes, file existence, a regex, and an
+error-tool_result count. No agent, no LLM judge.
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+from tools.benchmark import scoring
+from tools.benchmark.harness import load_task
+
+
+def _seed_passing_pytest(target: Path) -> None:
+    target.mkdir(parents=True, exist_ok=True)
+    (target / "pyproject.toml").write_text(
+        '[project]\nname = "t"\nversion = "0"\nrequires-python = ">=3.11"\n'
+        '[dependency-groups]\ndev = ["pytest"]\n'
+    )
+
+
+class TestPytestCheck:
+    def test_passes_when_tests_green_and_files_exist(self, tmp_path: Path):
+        _seed_passing_pytest(tmp_path)
+        (tmp_path / "slug.py").write_text("def slugify(s):\n    return s.lower()\n")
+        (tmp_path / "test_slug.py").write_text(
+            "from slug import slugify\n\ndef test_basic():\n    assert slugify('AB') == 'ab'\n"
+        )
+        assert scoring.score_check(load_task("feat"), tmp_path) is True
+
+    def test_fails_when_test_fails(self, tmp_path: Path):
+        _seed_passing_pytest(tmp_path)
+        (tmp_path / "slug.py").write_text("def slugify(s):\n    return s\n")  # wrong
+        (tmp_path / "test_slug.py").write_text(
+            "from slug import slugify\n\ndef test_basic():\n    assert slugify('AB') == 'ab'\n"
+        )
+        assert scoring.score_check(load_task("feat"), tmp_path) is False
+
+    def test_fails_when_required_file_missing(self, tmp_path: Path):
+        _seed_passing_pytest(tmp_path)
+        # must_exist names slug.py + test_slug.py; neither created → fail without
+        # even running pytest.
+        assert scoring.score_check(load_task("feat"), tmp_path) is False
+
+
+class TestRegexCheck:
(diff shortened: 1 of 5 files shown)


VytCepas:
**Review response** (Codex P1 — valid, fixed):

- **[scoring.py:54] honor fail_to_pass/pass_to_pass node ids**: Fixing — `_run_pytest_check` now runs the **explicit node ids** when the check lists `fail_to_pass`/`pass_to_pass` (new `_pytest_argv` keeps the runner + flags, drops the bare file arg, appends the node ids). Exit 0 then means the previously-failing node passes **and** every named passing node stayed green; a deleted/renamed node makes pytest exit non-zero → `success=False`. Tasks without node lists (`feat`) run the command as authored.

Added integration tests proving the integrity: correct fix → True; **deleting `test_divide_by_zero` → False** (the exact evasion you described — aggregate pytest would exit 0); breaking the pass_to_pass node → False; plus `_pytest_argv` unit tests.

`just lint && just test` green (1206 passed, 3 skipped).

VytCepas:
**Review response** (Copilot — both valid, both fixed; Codex's P1 was already addressed in the prior commit and re-anchored as stale):

- **[scoring.py:76] `uv run` can FileNotFoundError when uv isn't on PATH**: Fixing — added `_resolve_uv()` (mirrors `tests/helpers.find_uv`): resolves a leading `uv` via PATH then the common install locations before the subprocess, so the scorer can't emit a false-negative `success` just because `uv run` stripped uv from the child PATH.
- **[harness.py:212] non-string `result` would crash `re.search`**: Fixing — `run_task` now coerces a non-string `result` payload to `""` (`isinstance` guard) before it reaches the regex check.

Tests added for both. `just lint && just test` green (1208 passed, 3 skipped).

PR #453, opened 2026-06-24
chore(PI-452): benchmark harness supports Claude subscription auth

### LINKED ISSUE #452: Benchmark harness: support Claude subscription auth (seed credentials) + clear precondition error
## Problem

`tools/benchmark/harness.py` `run` isolates `CLAUDE_CONFIG_DIR` to an empty temp dir (harness.py:199, :321). That discards the user's OAuth/subscription credentials (`~/.claude/.credentials.json`), so `claude -p` returns `"Not logged in · Please run /login"` unless `ANTHROPIC_API_KEY` is exported. The run then **silently skips** every task (records the error's session_id, finds no transcript → "skipping") instead of failing clearly.

This blocks the whole study for the common case (Pro/Max subscription, no API key).

## Fix

1. Seed `.credentials.json` from the real config dir into the isolated config dir before running, so a **subscription** works without an API key (run stays otherwise isolated — clean fixed-overhead).
2. Add a clear precondition error when neither `ANTHROPIC_API_KEY` nor a credentials file is available, instead of the silent per-task skip.

## Out of scope

Collecting actual benchmark metrics (a separate manual run).

VytCepas:
## Problem

`tools/benchmark/harness.py run` isolated `CLAUDE_CONFIG_DIR` to an empty temp dir, discarding the user's OAuth/subscription credentials (`~/.claude/.credentials.json`). So `claude -p` returned `"Not logged in"` unless `ANTHROPIC_API_KEY` was exported — then the run *silently skipped* every task (recorded the error's session_id, found no transcript → "skipping"). This blocked the whole study for the common case (Pro/Max subscription, no API key).

## Fix

- `seed_credentials()` copies `.credentials.json` from the real config dir into the isolated run dir before launching, so a **subscription works without an API key**. The run stays otherwise isolated (clean fixed-overhead). API key still wins when set (env is inherited).
- `auth_available()` + an upfront precondition error in `run` replaces the silent per-task skip.
- `default_config_dir()` honors `CLAUDE_CONFIG_DIR`, else `~/.claude`.
- Token file copied at `0600` (never widens perms).

## Verification

- `uv run ruff check` — clean
- `uv run pytest` — 1258 passed, 3 skipped (6 new auth tests)
- **Ran for real on a Max subscription** (no API key): `harness run` now authenticates and writes records end-to-end.

Closes #452

### DIFF
--- a/tests/contracts/test_benchmark_harness.py
+++ b/tests/contracts/test_benchmark_harness.py
@@ -234,3 +234,46 @@ class TestRunTaskGuard:
         monkeypatch.setattr(harness.shutil, "which", lambda _: None)
         with pytest.raises(RuntimeError, match="claude.*CLI"):
             harness.run_task(tmp_path, {"prompt": "x"}, model="m", config_dir=tmp_path)
+
+
+class TestAuth:
+    """Subscription-or-API-key auth for the manual `run` path (no agent driven)."""
+
+    def test_default_config_dir_honors_env(self, tmp_path: Path, monkeypatch):
+        monkeypatch.setenv("CLAUDE_CONFIG_DIR", str(tmp_path / "cfg"))
+        assert harness.default_config_dir() == tmp_path / "cfg"
+        monkeypatch.delenv("CLAUDE_CONFIG_DIR", raising=False)
+        assert harness.default_config_dir() == Path.home() / ".claude"
+
+    def test_seed_credentials_copies_subscription_token(self, tmp_path: Path):
+        source = tmp_path / "real"
+        source.mkdir()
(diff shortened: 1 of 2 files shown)


VytCepas:
**Review response (cycle 1):** both addressed in af88804.
- [Codex P2] Honor OAuth-token auth — **Fixed.** `auth_available()` now accepts `CLAUDE_CODE_OAUTH_TOKEN` (documented non-interactive auth, precedence over stored creds), not just `ANTHROPIC_API_KEY`; added a test.
- [Copilot] chmod could widen a stricter source — **Fixed.** Now caps at the source's mode (`src.st_mode & 0o600`) so a 0o400 token isn't widened; added a regression test.

PR #455, opened 2026-06-24
chore(PI-454): verify current schema in add_hook/add_command; fix stale dev copy

VytCepas:
Closes #454

## What
- **`add_hook` / `add_command`**: add a best-effort **Step 0 — Confirm the current schema** that points the agent at Context7 / the live Claude Code docs to verify hook events and frontmatter fields before trusting the embedded snapshot. Falls back to the baseline when offline or `--no-egress`; never blocks.
- **Stale dev-copy fix**: this repo's own `.claude/skills/add_hook` still taught the deprecated `{"decision":"block","reason":...}` hook output schema. Corrected to the current `hookSpecificOutput`/`permissionDecision` form (kept `python3` — this repo has no `_py.sh`).

## Where
Canonical edit in `templates/fallback/dot_claude/skills/`, propagated byte-identical to the 6 surfaces (plugin, junie, codex, amp, antigravity) via `tools/sync_plugin.py`.

## Verification
- `ruff check` clean
- 1258 passed, 3 skipped
- 58 contract/drift-guard tests green (`test_plugin_marketplace`, `test_agent_overlays`, `test_agents_canonical`, `test_surface_emission`)

## Not included
Anthropic's built-in `skill-creator` (upstream-managed) and the global grill-me-codex / grill-with-docs-codex skills (edited separately, outside this repo).

PR #459, opened 2026-06-24
chore(PI-458): finish_pr verifies branch matches PR head before pushing

### LINKED ISSUE #458: finish_pr pushes the current branch without verifying it matches the PR head
Root cause (found during PI-456): `dag_workflow.py cmd_finish` calls `cmd_push(None, 3)` — pushing whatever branch is checked out — **before** resolving which PR it is finishing. If HEAD moves during the run (e.g. concurrent session activity), `finish` pushes the wrong branch. This happened during PR #455's merge: `finish_pr.sh 455` pushed an unrelated PI-450 branch.

Fix: in `cmd_finish`, resolve the PR first, read its head branch via `gh pr view <n> --json headRefName`, and refuse to push (fail-closed) if the checked-out branch does not match. Push by explicit branch name. `monitor_pr.sh` is already safe (operates by PR number).

Applies to all 3 byte-identical copies (base, plugin, dev).

VytCepas:
Closes #458

## Problem
`dag_workflow.py cmd_finish` pushed the **current** branch (`cmd_push(None, 3)`) *before* resolving which PR it was finishing. If HEAD moved mid-run (concurrent session activity), `finish` pushed the wrong branch. Observed during PR #455's merge: `finish_pr.sh 455` pushed an unrelated PI-450 branch.

## Fix
`cmd_finish` now:
1. Resolves the PR number first (arg or `_detect_pr_number`).
2. Reads its head branch via `gh pr view <n> --json headRefName`.
3. **Refuses to push (fail-closed) if the checked-out branch != PR head** — with a message telling the user to check out the right branch.
4. Pushes by explicit branch name.

`monitor_pr.sh` was already safe (operates purely by PR number); this closes the one roaming path.

## Scope
Applied identically to all 3 byte-identical copies (`templates/base`, plugin payload, dev `.claude/hooks`); synced + parity-checked.

## Tests
- `TestFinishBranchGuard`: mismatch → returns 1, **never** pushes; match → pushes the PR's head branch by name.
- Existing source/template-match contract still passes (copies identical).
- Full suite: 1260 passed, 3 skipped; ruff clean.

### DIFF
--- a/.claude/hooks/dag_workflow.py
+++ b/.claude/hooks/dag_workflow.py
@@ -500,6 +500,12 @@ def cmd_finish(pr_number: int | None, review_cycle: int | None) -> int:
         )
         return 1
     current = _current_branch()
+    if current is None:
+        sys.stderr.write(
+            f"finish: no current branch (detached HEAD or not a git repo) — check out "
+            f"PR #{pr_number}'s head '{head}' and re-run.\n"
+        )
+        return 1
     if current != head:
         sys.stderr.write(
             f"finish: checked-out branch '{current}' is not PR #{pr_number}'s head "
--- a/.claude/hooks/dag_workflow.py
+++ b/.claude/hooks/dag_workflow.py
@@ -482,14 +482,33 @@ def cmd_promote(pr_number: int | None) -> int:
 
 def cmd_finish(pr_number: int | None, review_cycle: int | None) -> int:
     """Push, promote, then hand off to monitor_pr.sh for CI/review/merge."""
-    rc = cmd_push(None, 3)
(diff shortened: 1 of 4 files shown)


VytCepas:
**Review response (Copilot, same comment ×3 copies):**

- **Detached-HEAD / `None` current branch: Fixing.** Added an explicit `current is None` branch that emits a clear message ('no current branch (detached HEAD or not a git repo) — check out PR #N's head ...') instead of the misleading `'None'`. Still fail-closed (no push). Added a regression test for it. Applied to all 3 copies + synced.

Issue #460, opened 2026-06-24
Add /live_test agent-driven live E2E command for the scaffolder


VytCepas:
Capture the manual 0.4.0 external live-test procedure (which found 13 bugs, epics #440–444) as a repeatable, user-only `/live_test` slash command living only in this repo.

The command builds the wheel once into an isolated dir, installs it into a throwaway venv, scaffolds a representative ~5 mini-project matrix (+1 interactive wizard smoke), fans out one isolated subagent per mini-project to exercise four depths (scaffolded gates, hooks/guards in the correct plugin/fallback mode, fully-local git/PR lifecycle with a stateful stubbed `gh` + bare remote, read-as-an-agent sanity), and aggregates a timestamped severity-ranked REPORT.md. No GitHub side effects, no auto-issue-filing, no merge path.

Plan was locked via grill + 2 Codex review rounds in a prior session.

Closes the e2e-test-command gap noted after the 0.4.0 live test.

PR #464, opened 2026-06-24
chore(PI-463): pin shell:bash on hooks + document Git for Windows + add Windows CI

### LINKED ISSUE #463: Native-Windows: document Git for Windows prerequisite + pin shell:bash + add Windows CI
Surfaced while scoping native-Windows testing for `/live_test` (#460). The scaffolder emits only bash hooks/scripts; this clarifies and locks the native-Windows (non-WSL) story.

## Finding
On native Windows, Claude Code runs hook commands via **Git Bash if present, else PowerShell** (official fallback chain). Our scaffolded hooks use POSIX `$CLAUDE_PROJECT_DIR`/`${CLAUDE_PLUGIN_ROOT}` expansion + bash `.sh` scripts. So:
- **With Git for Windows installed → works** (Claude Code auto-uses Git Bash; git runs its own hooks via bundled `sh`; `just` shells to `sh`).
- **Without it → hooks fail** (PowerShell can't run a bash shebang and won't expand `$VAR`).

The supported path (install Git for Windows, work from Git Bash) is **real but undocumented**, and nothing signals bash-intent on the hook objects.

## Resolution (design-aligned; keeps the single bash-3.2 floor — no .ps1 rewrite)
1. **Document Git for Windows as the native-Windows prerequisite** (README + using-project-init guide).
2. **Pin `"shell": "bash"`** on every command hook object in `templates/base/dot_claude/settings.json.tmpl` and `plugins/project-init-workflow/hooks/hooks.json` (officially supported; same schema in plugin hooks.json; already the default, so harmless on WSL/macOS/Linux — signals intent). + contract test.
3. **Add a gated `windows-latest` Git Bash CI job** mirroring `macos-portability` (scaffold `--no-plugin`, run portability checks) so the happy path can't silently regress.

## Out of scope
- PowerShell `.ps1` hook variants (contradicts the bash-floor ADR #359).
- Guaranteed loud-fail when bash is absent (Claude Code's behavior on a missing pinned shell is undocumented; the doc'd prerequisite covers it).

VytCepas:
## Why

Native Windows (non-WSL) runs Claude Code hooks through **Git Bash if present, else PowerShell** (official fallback chain). Our scaffolded hooks use POSIX `$CLAUDE_PROJECT_DIR` / `${CLAUDE_PLUGIN_ROOT}` expansion + bash `.sh` scripts — so without Git for Windows they silently no-op under PowerShell. The supported path (install Git for Windows, work from Git Bash) was **real but undocumented**, and nothing signalled bash-intent on the hook objects.

Surfaced while scoping native-Windows testing for `/live_test` (#460).

## What

- **Pin `"shell": "bash"`** on every command hook in `templates/base/dot_claude/settings.json.tmpl` and the plugin `plugins/project-init-workflow/hooks/hooks.json`. Officially supported; identical schema in plugin hooks.json; it's already the default so it's a no-op on WSL/macOS/Linux — it makes the Windows shell explicit and signals intent.
- **Contract test** (`test_hook_portability.py`) asserting every command hook carries the pin, across both the no-plugin `settings.json` and the plugin payload.
- **Docs**: Git for Windows as the native-Windows prerequisite (README platform note + `using-project-init` troubleshooting row). PowerShell-only stays a non-goal — no `.ps1` variants, per the bash-3.2 floor (ADR-017/#359).
- **`windows-latest` Git Bash CI job** mirroring `macos-portability` (gated to shell/template changes): scaffold `--no-plugin`, then run `session_setup` / `pre_commit_gate` / `_py.sh` under Git Bash so the happy path can't silently regress.

## Verification

- Full suite green (1263 passed, 3 skipped); `ruff check` clean.
- Rendered no-plugin `settings.json` has 6 valid `shell:bash` pins; plugin mode unchanged (hooks via plugin).
- New contract tests pass over both trees.

## Out of scope

- PowerShell `.ps1` hook variants (contradicts the single bash-floor ADR).
- Per-surface (codex/cursor/antigravity) `hooks.json` — different non-Claude schemas; not the `shell` field's domain.

Closes #463

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,3 +209,59 @@ jobs:
           cd /tmp/macos-target
           out=$(/bin/bash .claude/hooks/_py.sh -c "print('resolver-ok')")
           test "$out" = "resolver-ok"
+
+  # PI-463: native Windows (non-WSL) runs hooks through Git Bash when Git for
+  # Windows is present (the documented prerequisite). This is the Git Bash
+  # oracle — it proves the scaffolded bash hooks fire on Windows, catching
+  # CRLF / exec-bit / _py.sh-resolution breakage the Linux/macOS jobs can't.
+  # Gated to shell/template changes like the macOS job.
+  windows-portability:
+    name: Windows Git Bash portability
+    needs: changes
+    if: needs.changes.outputs.shell == 'true'
+    runs-on: windows-latest
+    defaults:
+      run:
+        shell: bash  # GitHub's "bash" on Windows is Git Bash — the supported path
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v8.1.0
+        with:
+          version: "0.11.7"
+          python-version: "3.12"
+
+      - name: Confirm bash is Git Bash
+        run: bash --version | head -1
+
+      - name: Scaffold a project
+        run: |
+          uv run project-init ./win-target \
+            --non-interactive --preset obsidian-only --name win \
+            --description "windows portability" --language python --no-plugin --strict
+
+      - name: SessionStart bootstrap runs under Git Bash
+        run: |
+          cd ./win-target
+          bash .claude/hooks/session_setup.sh
+          test -f .claude/.session_setup_stamp
+
+      - name: Commit gate runs under Git Bash against a staged file
+        run: |
+          cd ./win-target
+          git init -q
+          git config user.email ci@example.com
+          git config user.name ci
+          printf 'x = 1\n' > sample.py
+          git add sample.py
+          # PreToolUse hook: reads tool JSON on stdin, must exit 0 for a clean file.
+          printf '%s' '{"tool_input":{"command":"git commit -m test"}}' \
+            | bash .claude/hooks/pre_commit_gate.sh
+          echo "pre_commit_gate.sh exit: $?"
+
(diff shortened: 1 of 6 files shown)


VytCepas:

- **Codex P2 / Copilot (ci.yml — expand the `changes` path filter):** Already addressed in commit `61ec53a` (pushed after the reviewed `7ea73cf`). The filter now also matches `templates/**/settings.json.tmpl`, `plugins/**/hooks/**`, and `.github/workflows/ci.yml`, so hook-wiring changes (including this PR) trigger the macOS + Windows portability jobs. Verified live: both jobs ran on this PR.
- **Copilot (README — ADR-017):** Fixing. ADR-017 is the per-surface config generator, not the shell policy; no ADR covers the bash floor. Re-pointed to epic #359.
- **Copilot (ci.yml — exec-bit mention):** Fixing. The job invokes hooks via `bash <script>`, which doesn't validate exec bits. Corrected the comment; exec-bit coverage stays on Linux via `test_packaging`.
- **Also self-fixed:** the Windows job initially failed because it scaffolded into `./win-target` *inside* the checkout, so `session_setup.sh`'s `git rev-parse --show-toplevel` root resolution climbed to the parent repo. Now scaffolds into `$RUNNER_TEMP` with `CLAUDE_PROJECT_DIR` set explicitly (how Claude Code actually fires hooks), mirroring the macOS job.

PR #465, opened 2026-06-24
chore(PI-462): make plugin-mode scaffold docs mode-aware + fix ADR path

### LINKED ISSUE #462: Scaffolded-doc coherence bugs in default (plugin) mode — surfaced by /live_test (PI-460)
Found by running the new `/live_test` E2E harness (#460, PR #461) against a 5-preset matrix. The scaffolder **mechanics** are sound; these are **documentation-coherence** bugs where instruction files describe the `--no-plugin` wiring but ship in the **default (plugin)** scaffold. Two independent read-as-an-agent reviewers + a mechanical check agree on F1.

### F1 — P2 · Plugin-mode hook docs point agents at the wrong file
Scaffolded `AGENTS.md` states *"Hooks in `.claude/settings.json` fire automatically: `pre_commit_gate`, `github_command_guard`, `post_edit_lint`, `workflow_state_reminder`."* In the default plugin scaffold, `settings.json` has **no `hooks` block** — those hooks are supplied by the enabled `project-init-workflow@project-init` plugin. The line is accurate only for `--no-plugin` scaffolds.
- Repro: `project-init --preset governed --language python …` → `python3 -c "import json;print('hooks' in json.load(open('.claude/settings.json')))"` → `False`, yet AGENTS.md claims they live there.
- Source: `templates/base/AGENTS.md.tmpl:53` (emitted unconditionally; same ambiguity at `:43` prod_guard mention and `templates/base/dot_claude/rules/hooks.md`).
- Fix: gate on plugin mode — `{{#if no_plugin}}…settings.json…{{/if}}` + an `{{#unless no_plugin}}` variant pointing at the plugin (`enabledPlugins`; wired scripts under `CLAUDE_PLUGIN_ROOT`/hooks).

### F2 — P2 · Cursor is told to load a skill it can't reach
`AGENTS.md` instructs every agent to "load the `github_workflow` skill" before GitHub actions, but `.agents/skills/` is scoped to Codex+Antigravity (line 45), Cursor gets no `.cursor/skills/`, and can't load Claude plugin skills. Mitigated by the inline quick-ref on the same line.
- Fix: point Cursor at `.agents/skills/` (or a `.cursor/rules/*.mdc` pointer), or scope the instruction to surfaces that have a skills source.

### F3 — P3 · Scaffolded `.claude/skills/README.md` lists non-local skills
Documents `session_summary`/`start_task`/`add_hook`/`add_command` as local, but in plugin mode the only local skill is `plan`; the rest are plugin-provided. Make the table mode-aware.

### F4 — P3 · ADR path inconsistency
`CLAUDE.md:16` references `docs/adr/` but ADRs live in `.claude/docs/adr/` (`docs/adr/` doesn't exist). An agent could create a stray dir. Align the compact reference.

---
_Not a regression in scaffold mechanics. Split into separate PRs if preferred; F1 is the meaningful one._

VytCepas:
## Why

The `/live_test` run (#460) surfaced four doc-coherence bugs in scaffolded projects. They share **one root cause**: the instruction files describe the `--no-plugin` wiring, but the **default (plugin) scaffold** sources hooks *and* skills from the `project-init-workflow` plugin, not local files. No behavior change here — documentation accuracy only.

## What (F1–F4 from #462)

- **F1** — `AGENTS.md` / `project-init.md` said hooks fire "in `.claude/settings.json`". True only for `--no-plugin`; in plugin mode there's no `hooks` block (they're plugin-supplied). Split into mode-aware branches via the existing `{{#if no_plugin}}` / `{{#if plugin_mode}}` vars, so a hook-debugger is sent to the right place.
- **F2** — `AGENTS.md` told every surface to "load the `github_workflow` skill", but Cursor/VS Code have no skills dir and plugin-mode skills aren't in `.claude/skills/`. Added the portable-fallback note: surfaces that can't load skill files follow the inline quick-references (already present on the workflow line).
- **F3** — `.claude/skills/README.md` listed plugin-provided skills as local. Now a `.tmpl`, mode-aware (`plan` is local in both modes; the rest are plugin-provided by default).
- **F4** — `docs/adr/` references pointed at a path that **doesn't exist** in the scaffold; ADRs live in `.claude/docs/adr/` (where the `add_adr` skill writes and the scaffold creates them). Fixed all 9 references across 5 files.

## Locked by tests

`tests/contracts/test_agents_canonical.py`:
- `TestPluginModeHookDocs` — plugin-mode `AGENTS.md` must attribute hooks to the plugin and must **not** say "wired in settings.json"; `--no-plugin` must say the opposite.
- `TestAdrPathCanonical` — no instruction file may reference a bare `docs/adr/`; the canonical `.claude/docs/adr/` dir must exist.

## Verification

- Full suite green (1266 passed, 3 skipped); `ruff check` clean.
- Rendered both modes: plugin `AGENTS.md` attributes hooks to the plugin; `--no-plugin` to settings.json; F3 README correct in each; no scaffolded output references bare `docs/adr/`.

Closes #462


VytCepas:

- **Codex P2 / Copilot (project-init.md:16 — broken ADR link):** Fixed — and a real regression I introduced, thanks. `project-init.md` renders to `.claude/project-init.md`, so the `(.claude/docs/adr/)` href resolved to `.claude/.claude/docs/adr/`. Now `[`.claude/docs/adr/`](docs/adr/)` — displayed path absolute, href document-relative (resolves to `.claude/docs/adr/`). Verified against the rendered output.
- **Copilot (test regex misses `./` / `../`):** Fixed at the root — replaced the fragile bare-string ban with a **link-resolution** test (`test_adr_links_resolve`) that resolves every ADR href against its file's directory. That approach has no lookbehind blind spot and would have caught the broken link above.
- **Copilot (docs/README.md — ambiguous `vault/` paths):** Fixed — made the `vault/` and `adr/` cross-references root-absolute (`.claude/...`), consistent with the existing `.claude/vault/` reference. These are plain text (not links), so absolute is the root-correct form.

CLAUDE.md's ADR reference is plain text at repo root, so `.claude/docs/adr/` there is correct (not a link). Full suite green (1269 passed), `ruff` clean.

Issue #470, opened 2026-06-24
Decompose project-init into à-la-carte, self-explaining overlays


VytCepas:
## Vision

`project-init` should let a user **adopt exactly what they need — and understand why each piece is worth having — instead of swallowing the whole `base` layer at once.** Today `base` is a monolith: every scaffold force-ships an Obsidian memory stack, the full GitHub lifecycle (DAG hook + lifecycle scripts + board + wiki + templates), and a fixed toolchain bundle. The engine has *no* technical dependency on most of this — the coupling is prose links, `config.yaml` blocks, and preset `layers` lists.

The opt-in-overlay pattern is already proven here: `governance` (ADR-018), `observability` (ADR-019), per-surface `agents` (ADR-017), `graphify`, multi-model. Each is appended in `overlay_layers()` and gated by a wizard prompt that **explains its value before asking**. This epic extends that pattern to the concerns still trapped in `base`, and raises the wizard from yes/no to *"here is what this brings, and why."*

> This plan was hardened via cross-model review: a requirements grill + **3 rounds of adversarial Codex review**. That pass corrected a conceptual error in the memory-upgrade mechanism and caught two latent bugs before any code (see #466).

## Implementation order (LOCKED)

Dependency analysis: each workstream is **functionally independent** (each yields a working scaffold alone — no runtime dependency). The real coupling is **build-time file contention**: #466, B-impl, and C-impl all edit the same ~5 mechanism surfaces (`overlay_layers()` + its two call sites, the three variable-emit paths, the gated prose). Spikes are ADR/doc-only → no code collision. Therefore:

1. **#466 — WS-A Memory (impl).** FIRST. Spec-complete, lowest-risk, and **proves the overlay mechanism** that B-impl and C-impl reuse (resolved `ScaffoldInputs` field → `overlay_layers()` derivation → `{{#if}}` gate, identical across emit paths, round-trip-safe).
2. **#467 — WS-B lifecycle spike (ADR)** ‖ **#471 — WS-C toolchain spike (ADR).** After #466. **Parallel-able** — both are doc-only, no shared-file collision. Each decides its own boundary/default and emits an ADR.
3. **B-impl** (new issue, spawned from #467's ADR) — serialized impl. Then **C-impl** (new issue, spawned from #471's ADR) — serialized after B-impl (both touch the shared mechanism).
[...]

PR #473, opened 2026-06-24
feat(PI-466): make memory backend à-la-carte (memory=none, core preset)

### LINKED ISSUE #466: Make memory backend à-la-carte: add memory=none, decouple Obsidian from base
## Context

The scaffolder is layer-based, but the `base` layer is a monolith that **always** ships an Obsidian-shaped memory stack (`.claude/vault/` + `.claude/memory/`). Every shipped preset lists `obsidian` in `layers`, and `base` itself carries the vault stub — so "no Obsidian" is currently impossible.

Investigation confirms there is **no technical reason** the vault must be mandatory. The engine (`scaffold.py`) has zero dependency on it; the coupling is only:
- prose links in `AGENTS.md.tmpl` / `project-init.md.tmpl`
- the `memory:` block in `config.yaml.tmpl`
- `adr-001-memory-stack.md.tmpl`
- the `session_summary` row in `skills/README.md`
- `{{#if graphify}}` in `dot_gitignore.tmpl`

Memory should become a real **custom selection** (`obsidian-only` / `obsidian-graphify` / `none`), following the proven opt-in-overlay pattern already used by `governance`/`observability`.

## Approach (mirrors the `governance` overlay; PI-189 round-trip contract)

1. **Move memory content out of `base` into the `obsidian` layer** (no content change): `vault/**`, `memory/**`, `using-memory.md`, `adr-001-memory-stack.md.tmpl`. This is **invisible to `upgrade`** for existing obsidian projects — same output paths + hashes, so the manifest is unchanged. Only the new `none` path differs.
2. **Derive memory layers in `overlay_layers()`** via a new `memory` param: `obsidian-only → ["obsidian"]`, `obsidian-graphify → ["obsidian","graphify"]`, `none → []`. Drop `obsidian`/`graphify` from preset `layers` (presets keep the `memory_stack` var). Add a new vault-free **`core`** preset (`layers=["base"]`, `memory_stack="none"`) — the first preset without Obsidian.
3. **Wizard/CLI/variables** (the 11-touchpoint pattern): add `--memory none|obsidian|obsidian-graphify`, `memory` field on `ScaffoldInputs`, interactive prompt, flag-or-preset-var precedence (flag wins), new gating var `memory` (`""` when none). Pass `memory=...` to **both** `overlay_layers()` call sites (`main()` and `upgrade._render_staging()`) identically.
4. **Backfill** `memory_stack="obsidian-only"` / `memory="true"` in `_overlay_off_defaults()` so pre-existing records re-render faithfully.
5. **Gate the vault prose** behind `{{#if memory}}` in `AGENTS.md.tmpl`, `project-init.md.tmpl`, `config.yaml.tmpl`, `skills/README.md`.

## Tests

- New `tests/contracts/test_memory_none.py`: scaffold `core` → assert no `.claude/vault`/`.claude/memory`, no vault links in `AGENTS.md`, strict mode passes, core files still present.
- `overlay_layers` unit cases for the memory→layer mapping + stable ordering with agents/fallback/governance.
- Guard the memory-dependent assertions in `test_scaffold_obsidian.py` / `test_memory_starters.py` to obsidian backends only.
- Add `core` (+ a `none`-via-flag case) to the upgrade round-trip parametrization in `test_upgrade.py`.

## Docs

- README: document `--memory` + the `core` preset; correct the "Obsidian is mandatory" framing in *Positioning*.
- New ADR recording the memory-backend decomposition.

## Acceptance

- A vault-free scaffold is possible (`--preset core` or `--memory none`).
- `obsidian-only` / `obsidian-graphify` render byte-identically to today (manifest unchanged).
- `upgrade` round-trips with no drift for all three backends.
- Full suite green.

## Out of scope

- Decomposing the GitHub-workflow and toolchain concerns out of `base` (the GitHub one is tracked separately).

---

## ⚠️ Mechanism corrections & added requirements (grill + Codex review, 3 rounds)

The approach above is sound in shape but the **upgrade mechanism in step 4 was wrong**, and the reference/test inventory was incomplete. The corrections below **supersede** the conflicting parts. All paths are under `src/project_init/`.

### Corrected upgrade mechanism (memory is default-ON for existing records — NOT a default-off overlay)
- **Derive the memory overlay from the recorded `memory_stack`**, do **not** use `_overlay_off_defaults()` (that helper is for default-OFF features). `_backfill_variables()` (`upgrade.py:533`) and `_migrate_semantic_config()` (`upgrade.py:484`) already fall back `memory_stack→"obsidian-only"` for legacy records, so pre-feature records re-derive `["obsidian"]` with no new off-default entry. The gate is **computed from `memory_stack` every render**, never `setdefault`'d.
- `overlay_layers(..., memory_stack: str = "none")` — the default **must** be `"none"`. A default of `"obsidian-only"` would make `agent_layers()` (`__main__.py:955-957`) and existing tests that assert `overlay_layers("claude", no_plugin=False) == []` / `== ["codex"]` suddenly return `["obsidian"]`. The full scaffold (`__main__.py:1383`) and upgrade (`upgrade.py:628` `_render_staging`) call sites pass the **resolved** stack explicitly.
- **Bug fix — stack→preset-name collision:** `_migrate_semantic_config()` does `preset_name = stack  # 1:1` (`upgrade.py:484-485`). With `memory.stack: none` this calls `load_preset("none")`, which fails — the vault-free preset is named **`core`**. Map `none→core` (or decouple preset-name from stack in migration) and define legacy-migration behavior for `none` explicitly ("record-required; no pre-record legacy migration for `none`").
- **Explicit preset migration:** remove `obsidian`/`graphify` from preset `layers` in `templates/presets/*.toml` (`obsidian-only.toml` is `["base","obsidian"]` today). Memory must be represented ONLY by `memory_stack` + derived overlays — otherwise `--memory none` on a preset whose `layers` still lists `obsidian` still renders the vault.

### Rendered-variable contract (emit identically from all THREE paths: `_build_variables`, `_backfill_variables`, `_migrate_semantic_config`)
| `memory_stack` | `memory` (new gate) | `obsidian` | `graphify` |
|---|---|---|---|
| `none` | `""` | `""` | `""` |
| `obsidian-only` | `"true"` | `"true"` | `""` |
| `obsidian-graphify` | `"true"` | `"true"` | `"true"` |

The `{{#if memory}}` gate var is `memory`, distinct from `obsidian`; derived as `"" if memory_stack == "none" else "true"`. `ScaffoldInputs` gains a resolved `memory` field; `_build_variables()` must use it (not preset name/layers at `__main__.py:1122-1124,1161,1239-1240`). graphify ALWAYS implies obsidian; reject impossible combos (graphify without vault). Precedence: `--memory` flag > interactive > preset var > default. Define the full matrix incl. `--preset governed --memory none` and assert it.

### Expanded gating inventory (the original 5-file list was incomplete)
Gate behind `{{#if memory}}`, never delete: `AGENTS.md.tmpl`, `project-init.md.tmpl` (incl. `/session_summary`→vault), `config.yaml.tmpl` (`memory:` block), `skills/README.md`, **`docs/README.md`** (`.claude/vault`), **`dot_github/copilot-instructions.md.tmpl`** (`.claude/vault`+`.claude/memory`), **`dot_gitignore.tmpl`** — beyond the `{{#if graphify}}` block, the currently **UNCONDITIONAL** lines `.claude/vault/.obsidian/workspace*`, `.claude/vault/.obsidian/cache`, `.claude/vault/.trash/` (24-26) and `.claude/memory/.lightrag/`, `.claude/memory/.cache/` (6-7), and **`capabilities.py`** `_chosen_options` (reports "Memory stack" unconditionally, `capabilities.py:128-143`) + generated `CAPABILITIES.md` via `_emit_generated_files()`.

### File-level move table (single table; nothing stranded)
Define an exhaustive move table for every memory/vault file AND dependent scripts. `templates/obsidian/` already holds `lint_memory.sh` + vault Obsidian config; co-locate the `memory/SCHEMA.md` + starters it assumes (don't strand one without the other).

### `_default_preset_index()`
Adding `core` changes the visible preset list; update + test `_default_preset_index()` so Enter still selects the intended default.

### Tests (corrected & expanded)
- **Fresh-scaffold tree snapshot** for `obsidian-only`/`obsidian-graphify` against a **committed pre-move fixture** (path+hash manifest captured BEFORE the move) — scaffold-after-move-vs-itself is vacuous, and `memory/`+`vault/` are excluded from the upgrade manifest (`_PRESERVE_DIRS` `scaffold.py:35`), so upgrade round-trip alone won't cover the move.
- **Old-record upgrade**: recorded obsidian project (no `memory` field) upgrades with zero drift; compare full config **record** AND visible `config.yaml` body.
- `tests/contracts/test_memory_none.py`: scaffold `core` → no vault/memory dirs; **zero dangling links** to missing memory/vault across ALL rendered files (rg-allowlist); strict mode passes; `CAPABILITIES.md` shows no memory stack.
- `overlay_layers` mapping units (graphify⇒obsidian, none⇒[]); precedence-matrix units incl. governed; `_default_preset_index()` test; all-three-emit-paths variable-contract test.
- **Preserved-path exclusion** test (NOT consent-grouping — that was a false mechanism; `compute_drift()` skips `_is_preserved` before `_classify_addition()`, `upgrade.py:674-680`): assert memory/vault are intentionally excluded from `--accept-new/--decline-new`.

### New risks
- **R7 — Backend-switching via upgrade is undefined.** Because memory/vault are preserved, `upgrade` won't add/remove them when `memory_stack` changes. Decide whether switching backends is a supported upgrade path or re-scaffold-only; document.
- **R8 — Variable contract must be byte-identical across all three emit paths** or scaffold and upgrade diverge.

### Scope honesty
`core` = **"no memory backend", NOT "minimal"** — after this issue alone it still ships the full lifecycle + toolchain. Independence is narrowed to memory-only correctness. True minimalism emerges only once #467/WS-C land.

> Hardened via cross-model plan review (grill + 3 rounds of adversarial Codex). Part of epic #470 — implement **first** (foundation; proves the overlay mechanism #467-impl and the toolchain spike reuse).

VytCepas:
## What

WS-A of epic #470: make the memory backend **à-la-carte**. The Obsidian memory stack was force-shipped by `base`; now it's a declinable, self-explaining choice.

- New `--memory none|obsidian|obsidian-graphify` flag + vault-free **`core`** preset.
- Wizard renders a `rich.Panel` explaining each backend before asking (seeds the WS-D self-explaining-wizard pattern).
- Memory content (`memory/**`, `vault/**`, `using-memory.md`, scaffolded `adr-001`) moved `base → obsidian` overlay.

## How

- `overlay_layers(memory_stack=…)` derives the obsidian/graphify layers (default `"none"` so memory-agnostic callers like `agent_layers()` and existing tests are unaffected); passed explicitly at **both** call sites (scaffold + `upgrade._render_staging`).
- Presets drop `obsidian`/`graphify` from `layers`, keeping only the `memory_stack` var.
- One **variable contract** (`memory`/`obsidian`/`graphify`) emitted identically by `_build_variables`, `_backfill_variables`, and `_migrate_semantic_config`.
- Bug fix: `_migrate_semantic_config` mapped the stack 1:1 to a preset name → `load_preset("none")` would fail; `none` now maps to the `core` preset.
- Memory prose/config/`.gitignore` gated behind `{{#if memory}}` — never deleted.

## Round-trip safety (PI-189)

`obsidian-only` / `obsidian-graphify` / `governed` render **byte-identically** to before. `memory/`+`vault/` are excluded from the upgrade manifest, so the move is invisible to upgrade round-trip — it's verified instead by **committed pre-move tree-snapshot fixtures** (`tests/fixtures/memory_baseline/`).

## Scope notes (see ADR-020)

- `core` = "no memory backend", **not** "minimal" — it still ships the full lifecycle/toolchain (later #470 workstreams).
- Shared skills that *mention* memory (`save_memory`, `session_summary`, `audit`, `status`) are **not** gated: plugin mode (the default) serves skills from a static payload that can't carry `{{#if}}`, so per-backend gating would be inconsistent. They degrade gracefully; `core`'s guarantee is no vault/memory **dirs**, no memory **config**, no dangling **links**.

## Tests

- `test_memory_byte_identity.py` — fresh-scaffold snapshots vs committed baseline (4 combos).
- `test_memory_none.py` — overlay/variable contract (3 emit paths), default-preset guard, vault-free `core` contract (no dirs/config/dangling links).
- `test_memory_backends.py` — CLI precedence (incl. `governed --memory none`), core upgrade round-trip, old-record (missing `memory` var) upgrade, preserved-path exclusion.
- Full suite green (1303 passed); `ruff check` clean.

Closes #466.


VytCepas:
**Review response:**
- [Copilot — `_migrate_semantic_config` defaults to obsidian-only with no memory section]: **Fixing.** Good catch — this was a real faithfulness gap in the record-stripped fallback path for `core`. Changed the default from `obsidian-only` to `none` (commit on this branch). Safe for backward-compat: every pre-#466 config wrote a `memory:` block, so an absent section uniquely identifies a post-#466 `core` project. Added a regression test (`test_migrate_no_memory_section_is_core_not_obsidian`).

VytCepas:
**Review response:**
- [Codex P2 — config.yaml body not synced when switching memory backends]: **Not applying in this PR; documented instead.** This is the general config-preservation behavior (PI-179), not memory-specific: `config.yaml` is user-owned, so once a scaffold record exists, re-running with any changed flag updates only the record block, never the visible body. ADR-020 already scopes backend-switching as a re-scaffold/record-only operation; I've expanded it to call out the config-body staleness explicitly. The **fresh** `core` scaffold — the supported path — renders no memory block and is tested. Targeted config-body sync on a changed backend is a broader config-migration concern (affects all flags, not just memory) and belongs in its own workstream, not WS-A.

Issue #477, opened 2026-06-24
Toolchain: add a docs axis + gate renovate.json [ADR-022]


VytCepas:
## Context

Implements **ADR-022** (the #471 spike). The toolchain is already à-la-carte via whole-file `{{#if}}` conditionals — the spike concluded **keep the conditionals, no new overlays**, with two small refinements. Part of epic #470 (WS-C impl).

## Scope (small, additive — no `overlay_layers()` change, no registry)

1. **Introduce a docs axis** (`--docs` / `want_docs`) decoupling docs tooling from raw language:
   - `mkdocs.yml.tmpl`: gate on `want_docs` instead of `{{#if python}}`.
   - `typedoc.json.tmpl`: keep `{{#if node}}` but additionally respect `want_docs`.
   - These are **local** configs (`mkdocs serve` / `typedoc`) — there is **no** published-site workflow (PI-343/ADR-004 retired `docs.yml`).
   - **Backward-compat:** `want_docs` defaults to preserve today's output (on for python/node) so existing scaffolds re-render byte-identically (PI-189); the new behavior is the opt-out + cross-language opt-in.
2. **Gate `renovate.json`** (currently a static always-on file) behind a flag (e.g. `--renovate`, default on) or convert to `.tmpl` wrapped in `{{#if renovate}}`.
3. **Cleanup:** fix the stale comment in `mkdocs.yml.tmpl:3` ("Published to GitHub Pages by `.github/workflows/docs.yml`") to describe local preview only.

## Mechanism
Thread `want_docs` + `renovate` through the standard 11-touchpoint wizard/CLI/variable pattern (flag, `ScaffoldInputs` field, prompt, precedence, gating var), emitted identically across the three variable-emit paths. DoD includes a wizard Panel for the docs choice (WS-D pattern).

## Acceptance
- A `--docs none` scaffold omits mkdocs/typedoc configs; a `--docs` opt-in works cross-language. `--renovate` gates renovate.json.
- Defaults preserve current output (byte-identical); `upgrade` zero drift; full suite green.

> Spawned from ADR-022. Part of epic #470. Run AFTER B-impl (shared mechanism collision).

Issue #476, opened 2026-06-24
Extract GitHub lifecycle into an opt-out overlay (--lifecycle) [ADR-021]


VytCepas:
## Context

Implements **ADR-021** (the #467 spike). Extract the force-bundled GitHub lifecycle out of `base` into a declinable **opt-out overlay**, keeping the forge-agnostic quality tier as non-declinable core. Part of epic #470 (WS-B impl). Reuses the memory-overlay mechanism from #466/ADR-020.

## Scope (per ADR-021's classification)

**Move to the lifecycle overlay (gated by a new `--lifecycle`/`lifecycle` var):**
- Hooks: `dag_workflow.py`, `github_command_guard.sh`, `workflow_state_reminder.sh`
- Scripts: `create_issue`, `start_issue`, `create_nojira_pr`, `promote_review`, `monitor_pr`, `finish_pr`, `push_branch`, `setup_github`, `push_wiki`, `gh_host`
- Workflows: `board-automation.yml`, `issue-validation.yml`, `review-status.yml`, `validate-pr.yml.tmpl`
- `ISSUE_TEMPLATE/**`, `pull_request_template.md`
- Skills: `create_issue`, `start_task`, `github_workflow`, `request_review`, **`audit`** (creates GitHub issues — gate or rewrite lifecycle-free)
- **Split `dot_github/hooks/pre-push`**: keep the main/master push block in quality core; move the `<type>/…` branch-prefix check to the lifecycle tier.
- **Gate `project-init-upgrade.yml.tmpl`** (uses `gh pr list/create`) — must not ship under `--lifecycle none`.

**Stays core (quality tier):** `ci.yml` (quality jobs), `commit-msg`, `pre-commit` (gitleaks), `pre-push` (main/master block), `install_hooks.sh`, `pre_commit_gate.sh`, `post_edit_lint.sh`, `prod_guard.py`.

## Mechanism (reuse #466)
- Resolved `ScaffoldInputs.lifecycle` field → `overlay_layers()` derivation → `{{#if lifecycle}}` gate; identical across `_build_variables`/`_backfill_variables`/`_migrate_semantic_config`.
- **Plugin/no-plugin duality** (ADR-021 finding #1): split the `{{#if no_plugin}}` settings.json hook block AND the `project-init-workflow` plugin payload — lifecycle hooks move out (likely a separate `project-init-lifecycle` plugin); quality/safety/base hooks stay.
[...]

PR #484, opened 2026-06-25
feat(PI-472): self-explaining wizard standard with a coverage test

### LINKED ISSUE #472: Self-explaining wizard standard: explain the value of every selectable concern
## Context

Part of epic #470 (decompose `project-init` into à-la-carte overlays). The epic's headline goal is not just *making* concerns optional, but **explaining the value of each choice in the wizard** so a user makes an informed decision instead of blindly accepting or declining. This issue makes "explain why" a **standard**, not an ad-hoc per-feature habit.

## The proven pattern (already in the codebase)

`_choose_observability_interactive()`, `_choose_governance_interactive()`, and `_choose_multi_model_interactive()` already render a `rich.Panel` with the shape: **what it ships · "Helps:" line · honest cost · safe default**. That is the template — this issue promotes it to a documented, enforced standard.

## Model: panel-per-feature DoD + a standardizing pass (decided in the grill)

1. **Each feature workstream ships its own Panel as Definition-of-Done** — #466 (`_choose_memory_interactive`), #467/B-impl, WS-C/C-impl each add their explanation when they land. Explanations ship *with* the feature, not in a deferred sweep.
2. **This issue (runs LAST) then:**
   - Writes the **explanation standard** as an ADR / CONTRIBUTING rule (the Panel shape + copy guidelines: what-it-is, what-it-brings, cost, default).
   - **Retrofits selections that have no Panel today** — notably `_choose_preset_interactive()` (and the `core`/base path) and any pre-existing selection picked with little guidance.
   - Adds a **coverage test** whose enumeration source is the concrete flag/overlay/preset registry (so it can't be vacuous), asserting **every selectable concern has an explanation**.

## Acceptance

- A documented standard exists for wizard explanations.
- Every interactive selection (preset, memory, lifecycle, toolchain, overlays) presents a clear "what this brings / why you'd want it / cost / default" before asking; non-interactive flags are documented equivalently.
- The coverage test fails if a new selectable concern is added without an explanation.
- `--yes`/non-interactive paths keep working unchanged.

## Out of scope

- The feature overlays themselves (#466, #467, WS-C) — they own their individual Panels.
- A heavyweight data-driven framework — reuse the existing `_choose_*_interactive` shape; this is a standard + retrofit + test, not an abstraction layer.

> Part of epic #470. Runs **last** (consolidation) — depends only on the set of selections existing, blocks nothing. Hardened via grill + Codex plan review.

VytCepas:
Closes epic #470 (decompose project-init into à-la-carte, self-explaining overlays) with **WS-D** — the final consolidation. Promotes the ad-hoc "explain the value before asking" habit to a documented, test-enforced standard.

Closes #472

## What changed

- **ADR-023 — the standard**: every selectable concern explains its value before asking, in one of two forms — a `rich.Panel` (what it ships · a `Helps:` line · the cost · the default) for opt-in/opt-out concerns, or an annotated `name — description` option list for multi-choice pickers. Non-interactive flags carry the same info in `--help`/README.
- **Retrofit** the selections that asked with no explanation: the `devcontainer`/`mise`/`vscode`/`docs`/`renovate` toggles now render a concise explanation (shared `_explain_and_confirm` helper), and the **preset chooser** gains a value-framing panel.
- **Non-vacuous coverage test** (`test_wizard_explanations.py`): the enumeration source is the real CLI flag registry — `WIZARD_CONCERN_FLAGS` (concern→flag) and `WIZARD_MECHANICAL_FLAGS` must *partition every flag* in `_build_parser()`, so a newly added flag fails the test until classified; and every registered concern's chooser is rendered (prompts mocked) and asserted to actually print a value cue (`Helps:` or an annotated ` — ` list).
- **CLAUDE.md** documents the rule for contributors.

## Notes

- Deliberately a convention + helper + test, **not** a data-driven framework (per #472's out-of-scope) — the bespoke panels stay readable.
- Non-interactive / `--yes` paths are unchanged (the explainers only run in the interactive wizard).
- Full suite: **1381 passed**, ruff clean.

## Epic #470 — complete after this

A (memory #466) · B (lifecycle #476) · C (toolchain #477) · **D (this)** — every force-bundled concern is now declinable *and* explained. A minimal, knowingly-chosen scaffold: `project-init --preset core --memory none --lifecycle none --no-docs --no-renovate`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -35,6 +35,7 @@ Template naming convention: directories stored as `dot_claude/`, `dot_gitignore`
 - **ruff only** — no black / isort / mypy.
 - **justfile is the command surface** — `just --list` shows the canonical recipes (`setup`, `lint`, `format`, `test`, `docs`, `ci`); prefer `just <recipe>` over raw tool invocations. Recipes stay thin wrappers — no logic in the justfile.
 - **Templates are tested by scaffolding into a temp dir** — any change to `templates/` should have a corresponding test in the focused `tests/test_*.py` module for that behavior. Create a new focused file if no existing module fits.
+- **Self-explaining wizard (ADR-023)** — every *selectable concern* the wizard offers must explain its value before asking (a `rich.Panel` with what·`Helps:`·cost·default, or an annotated `name — description` option list). When you add an optional concern, add its chooser, a `WIZARD_CONCERN_FLAGS` entry, and `--help`/README copy — `test_wizard_explanations.py` partitions every CLI flag into concern-with-explainer or mechanical and fails if a new flag is unclassified.
 
 ## Settings
 
--- a/docs/adr/adr-023-wizard-explanation-standard.md
+++ b/docs/adr/adr-023-wizard-explanation-standard.md
@@ -23,8 +23,8 @@ so the next selectable concern can't ship as a bare yes/no prompt.
 forms satisfy the standard:
 
 1. **Panel form** (opt-in / opt-out concerns: memory, lifecycle, governance,
-   observability, multi-model, devcontainer, mise, vscode, docs, renovate). A
-   `rich.Panel` whose body states, in order:
+   observability, multi-model, devcontainer, mise, vscode, docs, renovate, the
+   Playwright/browser add-on). A `rich.Panel` whose body states, in order:
    - **what it ships** — the concrete files/behavior;
    - a **`Helps:` line** — the value, in the user's terms ("why you'd want it");
    - the **honest cost / caveat** — build time, a setting to enable, a tradeoff;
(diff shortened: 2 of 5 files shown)


VytCepas:

- **Codex P2 + Copilot — `browser`/Playwright classified as mechanical**: Fixed (both of you caught the same gap, thanks). `--browser` adds the Playwright MCP, so it *is* a selectable concern — classifying it mechanical meant the coverage test skipped it, exactly the loophole the test exists to prevent. Now `_choose_browser_interactive()` renders the standard explanation (what it ships · `Helps:` · cost · default), `browser` moved to `WIZARD_CONCERN_FLAGS`, and it's in the test's render map. ADR-023's concern list updated.

Full suite: 1382 passed, ruff clean.

PR #485, opened 2026-06-25
test: add a combination-matrix smoke test

VytCepas:
A focused guardrail closing the one gap flagged when epic #470 (à-la-carte decomposition) landed: a green unit suite can pass while a *combination* of overlays is broken, because the per-feature tests each pin a single axis.

This drives the **real CLI** (`main()` in `--strict` mode — which returns non-zero on any unrendered `{{placeholder}}`) across the cross-product of the interacting decomposition axes:

- **memory** (none / obsidian / obsidian-graphify) × **lifecycle** (github / none) × **plugin-mode** (plugin / no-plugin) × **language** (python / node / go / none) = **48 combinations**, plus a **minimal** extreme (everything declinable off) and a **maximal** one (all overlays + toolchain on).

Each case asserts not just "renders clean" but that the **gating took effect** for that combo: vault present iff memory on, lifecycle scripts present iff lifecycle on, the `project-init-lifecycle` plugin enabled iff (lifecycle on AND plugin mode), and the docs config matching the language. So a future change that breaks one corner can't pass on a green-elsewhere suite.

50 cases, ~1.6s. No behavior change — test infrastructure only. Full suite: 1432 passed, ruff clean.

### DIFF
--- a/tests/integration/test_combination_matrix.py
+++ b/tests/integration/test_combination_matrix.py
@@ -1,0 +1,140 @@
+"""Combination-matrix smoke test for the à-la-carte decomposition (epic #470).
+
+A green unit suite can hide a broken *corner* of the configuration space: the
+overlays interact (memory × lifecycle × plugin-mode × language, plus the
+toolchain toggles), and the per-feature tests each pin one axis. This drives the
+real CLI (`main()` in `--strict` mode, which fails on any unrendered placeholder)
+across the cross-product of the decomposition axes and asserts each combination
+both renders cleanly AND gates the right files — so a future change that breaks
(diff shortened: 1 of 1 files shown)


VytCepas:
**Review response:**

- **Copilot — lifecycle-plugin assertion too loose**: Fixed. You're right — `(x is True) == expected` would pass for a present-as-`false` (or unexpectedly-present) key, contradicting the comment that says it should be absent. Now it asserts the documented behavior precisely: `enabled[key] is True` when (lifecycle on AND plugin mode), and `key not in enabled` otherwise.

50 cases pass, ruff clean.

PR #489, opened 2026-06-25
fix(PI-488): Guard OpenTofu (tofu) in prod_guard deny-table

### LINKED ISSUE #488: prod_guard.py deny-table misses OpenTofu (tofu) destroy/apply -destroy
Found by the `/live_test` E2E pass (2026-06-25, proj4 = `--iac opentofu` go service).

## Problem
`prod_guard.py`'s `DENY_RULES` only match the `terraform` binary:

```python
(re.compile(r"\bterraform\s+(destroy|apply\s+.*-destroy)\b"), "terraform destroy"),
```

But a project scaffolded with `--iac opentofu` uses the **`tofu`** binary exclusively (`infra/README.md`, `infra.yml`, `versions.tf`). So a local agent running `tofu destroy -auto-approve` or `tofu apply -destroy` is **not flagged** — the guard's single most-relevant rule for that project's chosen IaC tool never fires.

The gap is in both the source-of-truth template (`templates/base/dot_claude/hooks/prod_guard.py`) and, via `sync_plugin.py`, the plugin payload (`plugins/project-init-workflow/hooks/prod_guard.py`). `tofu` count = 0 in both.

## Fix
OpenTofu is CLI-identical to Terraform (it's a fork), so the correct fix is **parity**: cover `tofu` with the same rule that already covers `terraform`. Fold both into one alternation rule rather than catching `-auto-approve` broadly (that would change existing terraform behavior and risk false positives on routine non-interactive applies).

## Acceptance
- `tofu destroy -auto-approve` → flagged (ask interactive / deny autonomous), mirroring `terraform destroy`.
- `tofu plan` and routine `tofu apply -auto-approve` → still pass (parity with terraform).
- Regression test in `tests/contracts/test_prod_guard.py`.
- `sync_plugin.py` re-run so the plugin payload copy matches.

VytCepas:
## Summary
`prod_guard.py`'s `DENY_RULES` only matched the `terraform` binary. A project scaffolded with `--iac opentofu` uses the **`tofu`** binary exclusively, so its single most-relevant destructive command was unguarded: `tofu destroy -auto-approve` and `tofu apply -destroy` were never flagged.

OpenTofu is a CLI-identical Terraform fork, so this folds both into one alternation rule for **parity**. Routine non-interactive `apply -auto-approve` stays unflagged (matching existing terraform behavior — no new false positives on normal CI applies).

## Found by
The repeatable `/live_test` E2E pass (2026-06-25), proj4 = `--iac opentofu` go service. This was the only P2 across the 5-project matrix.

## Changes
- `templates/base/dot_claude/hooks/prod_guard.py` — single `(?:terraform|tofu)` rule (source of truth).
- `plugins/project-init-workflow/hooks/prod_guard.py` — synced via `tools/sync_plugin.py` (byte-identical to the template copy).
- `tests/contracts/test_prod_guard.py` — tofu parity cases: `tofu destroy -auto-approve` + `tofu apply -destroy` flagged; `tofu plan` + `tofu apply -auto-approve` pass.
- 8 lifecycle/memory byte-identity fixtures — updated **only** the `prod_guard.py` hash (the one verbatim-copied file that changed); every other hash untouched, which still proves no other rendered byte drifted.

## Verification
- `uv run pytest` → 1443 passed, 3 skipped.
- Real firing: `tofu destroy -auto-approve` → `deny` (autonomous); `tofu apply -auto-approve` → passes.

Closes #488

### DIFF
--- a/plugins/project-init-workflow/hooks/prod_guard.py
+++ b/plugins/project-init-workflow/hooks/prod_guard.py
@@ -36,7 +36,11 @@ from pathlib import Path
 # names, which the ask/allowlist paths absorb cheaply.
 _SEG = r"[^|;&]*?"
 DENY_RULES: list[tuple[re.Pattern[str], str]] = [
-    (re.compile(r"\bterraform\s+(destroy|apply\s+.*-destroy)\b"), "terraform destroy"),
+    # OpenTofu (`tofu`) is a CLI-identical Terraform fork — guard both the same
+    # way (PI-488). Routine non-interactive `apply -auto-approve` is intentionally
+    # not flagged; only destroy / apply-with-destroy is.
+    (re.compile(r"\b(?:terraform|tofu)\s+(destroy|apply\s+.*-destroy)\b"),
+     "terraform/tofu destroy"),
     (re.compile(rf"\bkubectl\b{_SEG}\bdelete\b"), "kubectl delete"),
     (re.compile(rf"\bhelm\b{_SEG}\b(uninstall|delete)\b"), "helm uninstall"),
     (re.compile(rf"\baws\b{_SEG}\b(delete|terminate|remove)\S*\b"), "aws delete/terminate"),
--- a/plugins/project-init-workflow/hooks/prod_guard.py
+++ b/plugins/project-init-workflow/hooks/prod_guard.py
@@ -37,10 +37,13 @@ from pathlib import Path
 _SEG = r"[^|;&]*?"
 DENY_RULES: list[tuple[re.Pattern[str], str]] = [
     # OpenTofu (`tofu`) is a CLI-identical Terraform fork — guard both the same
-    # way (PI-488). Routine non-interactive `apply -auto-approve` is intentionally
-    # not flagged; only destroy / apply-with-destroy is.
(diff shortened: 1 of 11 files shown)


VytCepas:

- **Codex (P2 — global options before the verb):** Fixing. Good catch — the rule only allowed whitespace before the destructive verb, so `tofu -chdir=infra destroy` slipped past, inconsistent with the other rules' `_SEG` flag-tolerance. Added `(?:-\S+\s+)*` to skip leading option tokens. I kept it flag-*specific* rather than reusing the looser `_SEG`, because `_SEG` would also match the read-only `tofu plan -destroy` (a preview, not a destroy) and flag it as destructive. New tests cover `tofu -chdir=infra destroy` (flagged) and `tofu plan -destroy` + `tofu -chdir=infra apply -auto-approve` (not flagged).

- **Copilot (label precision for `apply -destroy`):** Fixing. Updated the rule label to `terraform/tofu destroy/apply -destroy` so the user-facing deny/ask reason is accurate when the match is the `apply -destroy` form.

Both copies synced via `tools/sync_plugin.py`; byte-identity fixtures updated for the new `prod_guard.py` hash only. Full suite: 1449 passed, 3 skipped.

PR #536, opened 2026-06-27
fix(PI-535): Harden never-clobber + idempotence in emission layer

VytCepas:
## What

Batch of emission-layer edge cases from an external review pass (Antigravity + Codex). The core merge/manifest engine is sound; these close crashes and weak spots in the generated-file / surface emission paths.

| Fix | Severity | File |
|---|---|---|
| `surfaces.emit` compares **bytes not decoded text** — non-UTF-8 existing file no longer crashes the scaffold; CRLF file no longer read-normalizes to "identical" and silently drifts | P2 | `surfaces.py` |
| `capabilities`/`governance` emit **never-clobber a pre-existing user file on first scaffold** (`.new` sibling); still overwrite on re-runs/upgrades | P3 | `scaffold.py` (new `_emit_generated`), `capabilities.py`, `governance.py` |
| `.new` siblings **excluded from the scaffold manifest** → no more spurious `removed` drift next upgrade | P3 | `upgrade.py` |
| `read_preserve_globs` **tolerates a non-UTF-8 `config.yaml`** | P3 | `scaffold.py` |
| Input validation **rejects U+0085 / U+2028 / U+2029** line separators (slip past the C0 check, break `splitlines()` YAML parsing) | P3 | `__main__.py` |
| `canonical_hooks` **warns** on invalid rendered `settings.json` instead of silently emptying the hook inventory | P3 | `capabilities.py` |
| Differential test: difflib 3-way fallback **agrees with `git merge-file`** on clean merges (incl. duplicate-anchor lines) | test | `tests/` |

## Investigated, confirmed NOT bugs (no change)

- **`apply_concern` manifest "poisoning" → data loss (claimed P1)** — the cited code path does not exist. `apply_concern` calls the same `apply_drift`; merged files are excluded from the manifest (`upgrade.py:870`), so they can never be classified `changed` and never silently overwritten.
- **version / `declined_additions` regex "comment hijack"** — anchors require the key after only whitespace; a `#`-prefixed comment can't match.
- **endless "merged" re-flag** — intentional (`upgrade.py:867-869`); re-merges cleanly, idempotent.

## Deferred follow-up

- Strict-mode atomicity for generated files (`_emit_generated_files` runs after the strict commit) — architecturally invasive for near-zero benefit; tracked for later.

## Tests

`uv run pytest -n auto` → **1599 passed, 3 skipped**. New module `tests/contracts/test_emission_hardening.py` (16 tests) + extended Unicode-separator cases in `test_cli.py`.

Closes #535

https://claude.ai/code/session_01NehbQYZo1dWvcVYaSdArUG

### DIFF
--- a/src/project_init/__main__.py
+++ b/src/project_init/__main__.py
@@ -1961,11 +1961,17 @@ def _validate_text_inputs(inputs: ScaffoldInputs, parser: argparse.ArgumentParse
         if (
             '"' in value
             or "\\" in value
-            or any(ord(ch) < 0x20 or ord(ch) == 0x7F for ch in value)
+            # 0x85 (NEL), 0x2028 (LINE SEPARATOR), 0x2029 (PARAGRAPH SEPARATOR)
+            # are > 0x20 so they slip past the C0 check, but Python's
+            # str.splitlines() treats them as line breaks — they'd split a
+            # single-line YAML value mid-parse on a later upgrade (PI-535).
+            or any(
+                ord(ch) < 0x20 or ord(ch) in (0x7F, 0x85, 0x2028, 0x2029) for ch in value
+            )
         ):
             parser.error(
                 f"--{flag} must not contain double-quotes, backslashes, newlines, "
-                "or control characters (they corrupt the generated config.yaml)"
+                "or control/line-separator characters (they corrupt the generated config.yaml)"
             )
 
 
--- a/src/project_init/__main__.py
+++ b/src/project_init/__main__.py
@@ -1975,6 +1975,27 @@ def _validate_text_inputs(inputs: ScaffoldInputs, parser: argparse.ArgumentParse
             )
 
 
+def _validate_existing_config(target: Path, parser: argparse.ArgumentParser) -> None:
+    """Reject a pre-existing, undecodable ``.claude/config.yaml`` before any writes.
+
+    The scaffold/upgrade readers tolerate non-UTF-8 bytes (errors="ignore") so
+    they don't crash mid-run, but ``write_scaffold_record`` rewrites the config
+    with strict UTF-8 afterwards. Letting the scaffold proceed would fail *late*,
+    after files are written — a partial-write state. Decode-check up front so the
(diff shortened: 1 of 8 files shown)


VytCepas:
**Review response:**

- **Codex (P2) — preserve generated files while a pending `.new` exists** (`scaffold.py` `_emit_generated`): Fixing — correct catch, this was a real run-2 data-loss path. My guard only protected on `first_scaffold`, so a re-run before the user merged the parked `.new` would overwrite the original. Now protects when `first_scaffold OR _has_pending_sibling(dest)`, matching the template-file convention in `scaffold()`. Added a regression test (`test_generated_file_protected_while_new_sibling_pending`).
- **Copilot — literal U+2028/U+2029 in test params** (`test_cli.py`): Fixing — replaced the literal separators with explicit ` ` / ` ` escape sequences so the exact codepoints are unambiguous in diffs.

Full suite: 1600 passed, 3 skipped; ruff clean. Pushed in 703eb00.

VytCepas:
@codex review

VytCepas:

- **Codex (P2) — reject invalid config before emitting files** (`scaffold.py:385`): Fixing — correct, the `errors="ignore"` tolerance turned the crash into a *late* failure after partial writes, since `write_scaffold_record` still reads the config with strict UTF-8. Took your "fail before any writes" option: added `_validate_existing_config()` that decode-checks a pre-existing `.claude/config.yaml` up front (before `_ensure_target_dir`) and aborts cleanly with nothing written. Regression test asserts no artifacts land. The reader tolerance stays as defense-in-depth for the upgrade/concern paths on already-valid configs.

Full suite: 1601 passed, 3 skipped; ruff clean.

VytCepas:
@codex review

VytCepas:

- **Copilot — docstrings omit the pending-sibling case** (`capabilities.py`, `governance.py`): Fixing — docstring-only. Both `emit` docstrings now state that protection also applies on a later run while an unmerged `.new` sibling is pending, not just on first scaffold. No behavior change.

PR #538, opened 2026-06-27
fix(PI-537): Post-dogfooding hardening batch (8 bugs)

VytCepas:
Eight confirmed bugs from dogfooding a generated project. Verified against source; triage in #537.

### P2
| # | Fix |
|---|---|
| **#1** | Memory toggle (`add/remove memory`) now refreshes the **visible** `memory:` descriptor in `config.yaml` (tier/stack/paths), not just the hidden record — `apply_drift` splices the freshly-rendered block (`_sync_memory_block`). An orchestrator reads the visible descriptor. |
| **#2** | `pre_commit_gate` / `post_edit_lint`: silence **stdout** on the ruff `--fix`/`format` calls (`ruff --quiet` still prints diagnostics there) so nothing leaks before the JSON hook response. |
| **#4** | codex/cursor/antigravity guard commands resolve from the **git root** (`git rev-parse --show-toplevel`) instead of a bare relative path that exits 127 / fails open from a subdirectory. |
| **#5** | Lifecycle skills (`create_issue`/`start_task`/`github_workflow`/`request_review`/`audit`) are shipped into the agent layers **gated on `{{#if lifecycle}}`**, so `--lifecycle none` drops them from `.agents/skills` as it already does from `.claude/skills` (`sync_plugin._ship_gated_skill`). |

### P3
- **#3** ruff-format the scaffold-owned `prod_guard.py` / `usage_report.py` / `gen_code_map.py` so a day-one `ruff format --check .` is clean.
- **#6** De-version the Codex hook advisory in `AGENTS.md` / `developer-onboarding.md` (dropped the `0.138.0` pin).
- **#7** `main()` catches `OSError` around scaffold/record write → clean error instead of a raw `PermissionError` traceback on a read-only target.
- **#8** Reject a whitespace-only `--name`/`--description` (was truthy, slipped past the empty-string check).

### Notes
- Byte-identity fixtures re-pinned for the intended #2/#3/#6 content changes — **only** those keys, every other file verified unchanged.
- `.agents` lifecycle skills render **byte-identically** when `--lifecycle github`; agent-overlay/portability tests updated to unwrap the gated `.tmpl`.
- Regression tests added for every behavioral fix. `uv run pytest -n auto` → **1610 passed, 3 skipped**; ruff clean; `sync_plugin` idempotent.

### Investigated, NOT bugs
typedoc.json is JSONC by design · template injection accepted-but-not-evaluated · strict-mode placeholder rejection · empty-name rejection · gitleaks fail-open (documented, CI-backed) · `prod_guard` string-matching `echo 'rm -rf /'` (conservative by design — shell-parsing would risk missing obfuscations).

Closes #537

https://claude.ai/code/session_01NehbQYZo1dWvcVYaSdArUG


VytCepas:
**Review response:**

- **Copilot — `_ship_gated_skill` write without `newline="\n"`** (`sync_plugin.py`): Fixing — added `newline="\n"` so the synced agent-skill tree stays byte-identical across hosts (a Windows checkout can't introduce CRLF). No byte change to the committed LF files.
- **Copilot — broad `OSError` message always reads as a target-write problem** (`__main__.py`): Fixing — the handler now names the actual failing path (`e.filename`) and OS reason, so a missing/unreadable template reads differently from a read-only target. Test updated.

(Codex was rate-limited and didn't review this PR.) Full suite: 1610 passed, 3 skipped; ruff clean.

PR #598, opened 2026-07-04
fix: resolve full-codebase review findings and release v0.6.1

VytCepas:
## What & why

Full bug-hunt across the wizard CLI, scaffold/upgrade engine, and the lifecycle hooks/scripts (source + templates + plugin payloads), followed by a diff-level code review and an end-to-end persona simulation (14 distinct usage profiles: node service + deploy/IaC, go minimalist, rust library, JSON orchestrator, no-plugin, all agent surfaces, governed org, RAG tier, custom-preset authors, upgrade round-trips, edge paths) — all green.

**Scaffolder (src/):**
- `upgrade --apply` no longer resets a hand-set `memory.rag_endpoint` when splicing the derived `memory:` block (ADR-025/ADR-026)
- preset `[vars]` control keys (`governance`/`lifecycle`/`memory_stack`) no longer leak into the template-gate namespace: `remove governance` on a governed-preset project now converges (previously left a broken CI job and regenerated the AI-BOM forever), and a preset's `lifecycle = "none"` no longer renders every `{{#if lifecycle}}` block ON
- upgrade stamps the current plugin version instead of a dead `setdefault` shadowed by the `"0.1.0"` migration backfill
- wizard: `--agents` honored without `--non-interactive`; language/license prompts normalize case and re-prompt instead of silently coercing to `none`; `--mcps` no longer suppresses the browser concern prompt (ADR-023)
- a preset's `lifecycle` var is validated up front (a TOML `false` silently shipped the full lifecycle overlay)
- CAPABILITIES.md shows the url for HTTP-transport MCP servers; `add memory obsidian` accepts the documented alias

**Guard + lifecycle scripts (all copies in lockstep, plugin payload synced):**
- `dag_workflow.py`: quoted git global-option values (`git -c foo='a b' push origin main`) no longer bypass every git rule including the push-to-main block; per-call subprocess timeout capped below the hook budget (raised to 60s) so one hung `gh` call can't make the fail-open PreToolUse hook time out; the nojira "PR already exists" short-circuit requires the PR to be OPEN
- `start_issue.sh`: a nonexistent issue reports an error instead of dying silently under `set -e`
- `monitor_pr.sh`: root copy backports the `--admin` opt-in for BLOCKED merges and the order-independent flag parser; both copies surface bot review comments after an early activity break instead of a false timeout message
- `pre_commit_gate.sh`: probes `uv run ruff --version` so uv projects without ruff fail open instead of blocking commits with uv's spawn error
- `push_wiki.sh`: empty `--prune` array no longer aborts under `set -u` on bash 3.2–4.3; `create_issue.sh`: temp file cleaned on early return

**Release prep:** version 0.6.1 across `pyproject.toml` / `__init__.py` / `CITATION.cff` / README / `uv.lock`, a 0.6.1 migration note, plugin bumps (workflow 0.2.1, lifecycle 0.1.1).

## Related issue

No tracking issue (review-driven fix batch).

## Checklist

- [x] `uv run ruff check .` + `uv run pytest -n auto` pass locally (1911 passed, 8 skipped) — the recipes `just ci` wraps
- [x] Changes under `templates/` have matching tests (`tests/contracts/test_lifecycle_script_hardening.py`, extended `test_dag_workflow.py`, `test_capabilities.py`, `test_audit_fixes.py`; byte-identity baselines re-pinned per the documented exception procedure)
- [x] Docs / README updated (README release line, migration note)
- [x] PR title follows Conventional Commits (`fix: description`, no issue)

https://claude.ai/code/session_01YLXJGTrZdeFt9f69pVqX81

---
_Generated by [Claude Code](https://claude.ai/code/session_01YLXJGTrZdeFt9f69pVqX81)_

### DIFF
--- a/.claude/hooks/dag_workflow.py
+++ b/.claude/hooks/dag_workflow.py
@@ -62,8 +62,11 @@ else:
 
 
 def _run(cmd: list[str]) -> tuple[int, str]:
+    # Per-call timeout must stay well below the hook budget (settings.json
+    # "timeout": 60): a PreToolUse hook that exceeds its budget is killed and
+    # FAILS OPEN, so one hung gh call must never eat the whole allowance.
     try:
-        proc = subprocess.run(cmd, capture_output=True, text=True, timeout=15)  # noqa: S603 — argv list built from internal literals/validated state, never a shell string
+        proc = subprocess.run(cmd, capture_output=True, text=True, timeout=10)  # noqa: S603 — argv list built from internal literals/validated state, never a shell string
     except (subprocess.TimeoutExpired, FileNotFoundError):
         return 1, ""
     return proc.returncode, proc.stdout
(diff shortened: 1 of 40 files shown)


VytCepas:
**Review response** (fixed in 9e55221):

- `--review-cycle` shift-2 abort: Fixing — the parser now validates a value is present before `shift 2` and rejects non-numeric values with a clear error (a bare `--review-cycle` used to die under `set -e` with no message, and a non-numeric value would have broken the later `-ne`/`-ge` arithmetic tests). Applied to both the root copy and the template copy, which had the same pre-existing defect.
- `gh issue view` stderr discarded: Fixing — stderr is now captured to a temp file and printed in the failure branch, so an auth/network failure reads as itself rather than "issue not found". Kept out of stdout capture deliberately: `2>&1` would leak gh's error text into `ISSUE_TITLE` and defeat the empty-check.

---

PR #599, opened 2026-07-04
feat: make release.yml dispatchable with an explicit tag

VytCepas:
## What & why

Cutting v0.6.1 surfaced a release-engineering gap: a tag can't always be *pushed* — release automation (like this session) may sit behind a branch-scoped git proxy that rejects `refs/tags/*` pushes, and the GitHub API surface available to it has no tag-creation endpoint. The release workflow was tag-push-only, so the release couldn't be cut at all from such an environment.

This makes `release.yml` additionally dispatchable with a required `tag` input:

- the input is verified against the `pyproject.toml` version exactly like a pushed tag (`GITHUB_REF_NAME` path unchanged),
- the changelog is scoped with `--unreleased --tag <input>` (the tag doesn't exist yet at generation time),
- `softprops/action-gh-release` creates the tag at the checked-out HEAD via `tag_name`,
- the `publish-pypi` job is untouched (trusted publishing, same workflow identity).

The tag-push path behaves byte-for-byte as before; `tag_name` on that path resolves to `github.ref_name`, which is what the action already defaulted to.

Contract coverage: `test_release_engineering.py::TestReleaseWorkflow::test_dispatch_release_path` pins the dispatch trigger, the input-based verification, the tag creation, and the unreleased-span changelog scoping.

## Related issue

No tracking issue (release-engineering follow-up to #598).

## Checklist

- [x] `uv run ruff check .` + `uv run pytest -n auto` pass locally (1912 passed, 8 skipped)
- [x] No `templates/` changes in this PR
- [x] Docs / README: no user-facing flag or behavior change
- [x] PR title follows Conventional Commits (`feat: description`, no issue)

https://claude.ai/code/session_01YLXJGTrZdeFt9f69pVqX81

---
_Generated by [Claude Code](https://claude.ai/code/session_01YLXJGTrZdeFt9f69pVqX81)_

### DIFF
--- a/.claude/hooks/dag_workflow.py
+++ b/.claude/hooks/dag_workflow.py
@@ -62,8 +62,11 @@ else:
 
 
 def _run(cmd: list[str]) -> tuple[int, str]:
+    # Per-call timeout must stay well below the hook budget (settings.json
+    # "timeout": 60): a PreToolUse hook that exceeds its budget is killed and
+    # FAILS OPEN, so one hung gh call must never eat the whole allowance.
     try:
-        proc = subprocess.run(cmd, capture_output=True, text=True, timeout=15)  # noqa: S603 — argv list built from internal literals/validated state, never a shell string
+        proc = subprocess.run(cmd, capture_output=True, text=True, timeout=10)  # noqa: S603 — argv list built from internal literals/validated state, never a shell string
     except (subprocess.TimeoutExpired, FileNotFoundError):
         return 1, ""
     return proc.returncode, proc.stdout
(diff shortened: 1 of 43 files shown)


VytCepas:
**Review response** (addressed in 31e7ea5 — the dispatch path was redesigned, which resolves all seven comments at once):

**Codex P1 (pypi environment denies branch-ref runs): Fixing — this was the decisive finding.** ADR-011 does restrict the `pypi` environment to `v*` tags, so the single-workflow design could never publish from a dispatch on `main`, and the GITHUB_TOKEN-created tag can't retrigger the tag-push path. The redesign is two-stage: a new `tag-release.yml` verifies the requested tag against the pyproject version, creates it at HEAD, and chain-dispatches `release.yml` **on the tag ref** (`workflow_dispatch` API calls are the documented exception to GITHUB_TOKEN's no-retrigger rule). The release run therefore lives on a `v*` ref and the environment rule is satisfied.

**Codex P2 (shell injection via `inputs.tag`): Fixing** — in `tag-release.yml` the input reaches the shell exclusively via `env:` (`INPUT_TAG`), never inline `${{ }}` inside `run:`; the contract test now asserts no `run:` script contains any inline interpolation. `release.yml` no longer has inputs at all.

**Codex P2 (`target_commitish` for created tags): No longer applicable** — the release action never creates a tag anymore; `tag-release.yml` creates it explicitly at the checked-out HEAD with plain `git tag && git push`, and refuses to move an existing tag.

**Copilot ×3 (`inputs` context on tag-push runs): Not applying as described, but moot after the redesign** — the `inputs` context is empty (not an evaluation error) on non-dispatch events, so `${{ inputs.tag || ... }}` was well-defined; regardless, `release.yml` now contains no `inputs` references, which is strictly simpler.

**Copilot (contract test too broad): Fixing** — `test_dispatch_release_path` now pins the actual wiring: bare dispatch + no inputs in `release.yml`, and in `tag-release.yml` the env-only input passing, the `v$PKG_VERSION` guard, the tag push, and the chain-dispatch.

---

Issue #565, opened 2026-07-01
feat: add Semgrep to CI as semantic security backstop


VytCepas:
## Summary

CI runs gitleaks (secret scanning) but no semantic code analysis. Patterns like unsafe subprocess calls, SQL injection, and OWASP Top 10 issues are not caught automatically.

## Proposed solution

- Add a `semgrep` job to the scaffolded CI workflow
- Run with `p/python`, `p/secrets`, `p/owasp-top-ten` rulesets
- Use `--baseline-commit` in PR context to scan only the diff
- Non-blocking initially; promote to blocking after the false positive rate is calibrated
- Pairs with gitleaks: gitleaks = string patterns, semgrep = semantic/AST patterns

## Notes

- Semgrep is the "out-of-loop backstop" — catches what the hook layer misses and what agents rationalize away
- Use explicit rulesets (`p/python`) over `--config=auto` for reproducibility in CI
- This is the second-to-last line of defense before the subagent reviewer

## Acceptance criteria

- [ ] `semgrep` job added to `ci.yml.tmpl` with explicit rulesets (`p/python`, `p/secrets`, `p/owasp-top-ten`)
- [ ] `--baseline-commit` used to scope PR scans to the diff
- [ ] Job is non-blocking initially (documented promotion criteria to blocking)
- [ ] Contract test asserts the semgrep job renders in the scaffolded CI workflow

### Priority

medium

### Size

S

### Agent ready

Yes

PR #630, opened 2026-07-07
fix: pin scaffolded workflow actions to commit SHAs (#629)

VytCepas:
## Problem

A fresh `project-init` scaffold fails its own Semgrep security gate. The CI Semgrep job (#565) runs `--config p/owasp-top-ten`, which flags `yaml.github-actions.security.github-actions-mutable-action-tag` as **blocking** — yet every shipped workflow pinned its actions to mutable tags (`@v6`, `@v8.1.0`, …). Result: **22 blocking findings on a clean scaffold**, `semgrep scan --error` exits 1, and the Semgrep check is red on effectively every PR (the lifecycle `finish_pr.sh` then refuses to merge).

Reported in #629; hit downstream in `VytCepas/zarija` on its first PR.

## Fix

- **Pin all 77 action refs** across the shipped workflow templates (`base` + `lifecycle`) to full 40-char commit SHAs with a trailing `# vX` comment. This is self-consistent with the repo's own workflows, which already use `helpers:pinGitHubActionDigests`.
- **Renovate keeps them fresh**: the templates' regex `customManager` is updated from the tag pattern to the digest pattern (`currentDigest` + `currentValue` from the `# vX` comment) and broadened to also cover plain workflow `.yml` (e.g. `board-automation.yml`), not only `.yml.tmpl`.
- **Contract tests updated**: assertions that matched the mutable-tag strings now assert the SHA-pinned form; `project-init-upgrade.yml` is added to the byte-identity exclusion sets in `test_lifecycle_byte_identity` and `test_memory_byte_identity` (`ci.yml` and `board-automation.yml` were already excluded) as a deliberate content change, following the established pattern.

## Verification

- `just lint` clean, full suite **1961 passed, 8 skipped** (was 14 failing after the template change, all resolved).
- No mutable action tags remain in any shipped workflow template.

Closes #629

https://claude.ai/code/session_01RWWT5gbukTdjVmAowhnQtj


VytCepas:
**Review response (Codex P2 — "Keep matching unpinned template actions"):**

Valid concern — thanks. Addressed it in bebae01, but via a *deterministic contract test* rather than a second Renovate matcher, because that's the stronger guarantee here:

- **Root gap confirmed & closed.** The existing coverage test only scanned `*.yml.tmpl`, so plain `.yml` workflows (`board-automation.yml`) were unguarded. `test_custom_manager_covers_workflow_templates` now audits **every** external action ref across all workflow templates (both extensions), using the manager regex as the is-pinned oracle. Verified it fails on an injected `actions/cache@v5` in `board-automation.yml` and passes on the current fully-pinned tree.
- **Why a contract test, not a second custom-manager match.** `helpers:pinGitHubActionDigests` applies to the built-in `github-actions` manager, not to a regex custom manager — a second matcher for unpinned refs would let Renovate bump the *version* but not pin it to a digest, so a mutable `@vN` could still ship and fail the scaffold's Semgrep gate. The contract test rejects the unpinned ref at author time in *this* repo's CI, which is exactly where we want the failure — before it ever reaches a fresh scaffold.

Net: a new template action added with a mutable tag now fails CI here immediately, deterministically.

PR #634, opened 2026-07-08
fix(PI-627): scope & harden the .claude projection; resolve bug-review findings

VytCepas:
## Summary

Two strands of work from a bug-review pass:

1. **`.claude` / `.agents` consolidation (#627)** — the projection is reworked into a scoped, delete-aware, cross-OS-safe form, and this repo now dogfoods it so its own guard hooks load.
2. **Four independent bug-review findings** across the CLI, wizard, and lifecycle DAG.

Every change is verified — full suite **1982 passed**, lint clean — and the `.claude` behaviour is confirmed end-to-end against the real Claude Code CLI (v2.1.204).

---

## Part 1 — `.claude` config projection (Closes #627)

**Grounded in empirical verification, not docs.** Claude Code reads project config (`settings.json`, hooks, skills, commands, subagents) from **`.claude/` only** — a hook defined solely in `.agents/settings.json` never fires; the same hook under `.claude/` does. So the projection is load-bearing, not legacy.

The old `copytree(.agents, .claude, dirs_exist_ok=True)` was wrong on the counts #627 raised:

| # | Problem | Fix |
|---|---------|-----|
| 1 | **Split-brain** — `memory/`, `vault/`, `docs/adr/`, `governance/`, `config.yaml` duplicated into `.claude/`; a memory write or new ADR under canonical `.agents/` drifts from the stale `.claude/` copy | Project **only the config surface Claude reads** (`settings.json`, `skills/`, `commands/`, `agents/`, `rules/`); exclude state/descriptors + machinery (`hooks/`, `scripts/`). Denylist, so future config dirs still project — never silently omitted |
| 2 | Manifest doesn't track `.claude/**` | Resolved the issue's own way — **regenerated-and-disposable**, never hand-edited (documented in `CLAUDE.md`) |
| 3 | `copytree` never deletes → stale files linger | **Delete-aware** rebuild (clear then copy) — the two cannot diverge |

**Why plain files, not a symlink.** A committed symlink is not portable: git's default `core.symlinks=false` on **both macOS and Windows** materializes a checked-out symlink as a plain text file, so Claude Code sees `.claude` as a *file* and silently loads nothing. Plain files (git mode `100644`) restore identically on every OS. Machinery is safe to exclude because `settings.json` points hooks at `.agents/hooks/` and scaffolded skills call `.agents/scripts/` (templates carry **zero** `.claude/` references).

**Verified end-to-end:** in plugin and `--no-plugin` modes, `.claude/` carries no `memory/vault/docs/governance/config.yaml/hooks/scripts`, yet a headless `claude -p` run still loads config (hook fires, skill discovered); and a `git commit` → fresh `git clone` restores `.claude/` as a real dir (mode `100644`) that loads on a bare checkout.

**Repo dogfood.** This repo authored its config under `.agents/` with no `.claude/`, so its own `github_command_guard`/`workflow_state_reminder` hooks never fired in a Claude session (and `.agents/settings.json` still pointed at a stale `.claude/hooks/` path from the PI-606 rename). Added a committed `.claude/` mirror via `tools/sync_claude_dir.py` (`just sync-claude`, run by `just setup`), drift-tested like the `plugins/` payload. Confirmed live: the repo's hooks now load in a Claude Code session.

The full decision, empirical tables, and rejected alternatives (symlink, single-folder, gitignore+self-heal) are recorded in **ADR-027**.

## Part 2 — Bug-review findings

| Area | Bug | Fix |
|------|-----|-----|
| Lifecycle DAG (`dag_workflow.py`) | `check_ci_green` ignored `StatusContext.state` — classic commit statuses (Vercel/Codecov/legacy CI) counted as "still running" forever, blocking the merge gate | Fold `state`; propagated to all three synced copies + fixtures re-pinned |
| Scaffold engine | A preset's `[vars] observability = true` turned the config.yaml gate ON while the overlay layer stayed absent | Add `observability` to `_PRESET_CONTROL_KEYS` |
| Wizard | Explicit `--agents claude` was indistinguishable from an absent flag → both fell through to the surface chooser (which can never return claude-only) | Default `--agents` to `None`; honor an explicit value |
| CLI descriptor | Python `run_command` used the kebab slug → `uv run python -m my-app` (invalid module name) | Substitute the underscore module form |

Each ships with a regression test.

## Testing

- **1982 passed**, 8 skipped; ruff clean.
- New: `test_claude_projection.py`, `test_claude_dir_sync.py`, plus regression tests for all four bug-review findings.
- Live verification against Claude Code CLI v2.1.204 (hook fire / skill discovery / git clone round-trip) documented in ADR-027.

## Decisions documented

- **ADR-027** — `.claude/` is a scoped, delete-aware, plain-file projection of `.agents/` (added to mkdocs nav).
- README, `surfaces.py`, and the non-CLI surface matrix corrected: Claude Code reads `.claude/` (a projection of `.agents/`), replacing the earlier "reads `.agents/` natively" claim.

Closes #627

https://claude.ai/code/session_01AtE9YMLmYX4rWyBwHtWZeh

---
_Generated by [Claude Code](https://claude.ai/code/session_01AtE9YMLmYX4rWyBwHtWZeh)_


VytCepas:
**Review response** — all comments addressed in `ef99213` + `0f4989e`:

**Substantive bugs (thanks — both real):**
- **P1** `scaffold.py:1078` (upgrade never projects to target): `upgrade --apply` renders into staging and applies only the `rendered` set, which excludes the derived projection. Fixed — the projection now re-runs against the target's updated `.agents/` after `apply_drift`. Test: `test_apply_regenerates_claude_projection_on_target`.
- **P2** `scaffold.py:1138` (adoption clobbered user config): the delete-aware rebuild only ever clears a projection *we* generated now. On the **first** scaffold, a non-empty pre-existing `.claude/` is parked as `.claude.pre-project-init` and reported as a conflict (PI-179 spirit), never deleted. Tests cover backup, unique-naming, empty-dir, and re-run cases.
- **`__main__.py:2121`** (run_command on upgrade): correct — the underscore module-name fix had only landed on the scaffold path. Centralized in a shared `render_run_command()` now used by the scaffold path and **both** upgrade backfill sites (`upgrade.py:600,723`), so `python -m my-app` can't reappear on upgrade.

**Pre-existing dev-infra content the mirror surfaced (all fixed in canonical `.agents/`, re-synced):**
- `push_wiki.sh`: added a bot git-identity fallback so the commit succeeds on CI runners without `user.name`/`email` (per-command `-c`, so a dev's global identity is untouched).
- `monitor_pr.sh`: header corrected — it does log periodic review-wait progress.
- `push_branch.sh`: comment now describes the relative `../hooks/dag_workflow.py` (not a hard-coded `.claude/` path).
- `preset-guide.md`: `tests/unit/test_presets.py` (+ inheritance file).
- `implementation-guide.md`: branch convention `<type>/PI-<n>-<kebab-slug>`.

ADR-027 records the upgrade-reprojection and adoption-safety behaviors. Full suite green (**1991 passed**).

---

VytCepas:
**Review response** (round 3) — addressed in `2befabc`:

- **`session_summary` → `.claude/vault/sessions/`** (good catch, real state-loss): `.claude/` is now a regenerated scoped projection and `vault/` is excluded state, so session notes written there would be wiped on the next sync. Repointed to the canonical, preserved `.agents/vault/sessions/`.
- **`scaffolder-logic.md`** (both comments): the wiki template still described `.claude/` as the primary output (`generates .claude/ layouts`, `dot_claude/ → .claude/`, `.claude/ is initialized with…`). Rewrote to the current model — `.agents/` canonical (`dot_agents/ → .agents/`) with a scoped `.claude/` projection of the Claude-read config surface (ADR-027); state/machinery stay in `.agents/` only.

Re-synced into `.claude/`; suite green (1991 passed). The earlier-round items (P1/P2, run_command, push_wiki/monitor_pr/push_branch, wiki test-path/branch-convention) are in `ef99213`/`0f4989e` — a few of those threads still show as open but the fixes are on the branch.

---

VytCepas:
Addressed in `d1997a4`: non-interactive resolution now only defaults to `claude` when `--agents` is **absent** (`None`); an explicit value — including `--agents ""` — is passed to `resolve_agents` (which validates it and still returns the always-included `["claude"]`). Keeps the absent-vs-explicit distinction consistent with the interactive path.

---

VytCepas:
**Review response** (round 4) — addressed in `694562f`:

- **`scaffold.py:1172` (symlink/file adoption)**: right — the first-scaffold guard only preserved a non-empty *dir*; a pre-existing `.claude` symlink or file was still unlinked. Now **any** pre-existing `.claude` on the first scaffold (dir, symlink, or file) is parked as `.claude.pre-project-init` and reported; only our own projection (later runs, or an empty dir) is replaced in place. Added symlink and file adoption tests.
- **`test_claude_projection.py:5` / scaffold docstring (hooks wording)**: clarified — Claude reads `settings.json` (hook *wiring*) + skills/commands/subagents from `.claude/`; hook *scripts* stay canonical in `.agents/hooks/` (settings.json points there), so `.claude/hooks/` is deliberately excluded. No more contradiction.
- **`test_audit_fixes.py:240` (fragile count)**: replaced the total-occurrence count with asserts on the two specific backfill assignments (+ the replace-guard), so unrelated future mentions of the function won't break it.

Suite green (**1994 passed**).

---

PR #642, opened 2026-07-08
chore: Add fast-iterate test recipe and token-efficiency conventions

VytCepas:
Quick-win subset of #641 (token-efficiency investigation). No behavior change to the guard/DAG.

- **`just test-quick`** — `pytest -x -q --tb=short` for the debug-iterate loop, so failing-test noise (tracebacks, fixture diffs) stays out of agent context. `just test` remains the full green check.
- **CLAUDE.md convention** — filter tool output *before* it enters the transcript (it's re-sent every turn): pipe/`tail`/`grep`, line-ranged reads, delegate broad searches to the `Explore` agent.

Part of #641. Does not close it — the reminder de-duplication, payload trim, watcher, and measurement items remain for the planning pass.


VytCepas:
**Review response:**
- CLAUDE.md:46 (Copilot) — Fixing: added `test-quick` to the canonical-recipes list so the command-surface doc matches the justfile.

PR #511, opened 2026-06-26
feat(PI-510): Add --json output for orchestrator-driven scaffolding

VytCepas:
## What

Machine-readable `--json` output so the **agentic-os orchestrator** (#509) can drive `project-init` as an external CLI cleanly — the project-init-side polish promised when finishing the descriptor contract (#498).

- **`--list-presets [--json]`** — discover presets (name / description / memory_stack) without needing `--name`/`--target`.
- **`--json` on a scaffold run** — emit a single JSON result (target, preset, the resolved memory descriptor incl. `tier`/paths/`rag_endpoint`, `contract_version`, `files_created`, `conflicts`) as the **sole stdout line**. The profile/egress notice and rich panels are suppressed so stdout stays machine-clean (verified by test).
- Both flags classified as mechanical in `WIZARD_CONCERN_FLAGS` (ADR-023 coverage test).
- Extracted `_emit_scaffold_output` / `_emit_preset_list` to keep `main()` under the complexity gate.

## Also — bug sweep result (#509)

This branch carries the only output from an adversarial bug sweep over the session's merged work (tier-3 RAG seam #506 + contract version #508). The sweep empirically scaffolded all 5 tiers in both modes and ran upgrade round-trips, and found **no correctness, consistency, or data-integrity bugs** — only two P3 doc-consistency nits, both fixed here:

- `config.yaml` tier/stack comment legend now includes tier 3 (`obsidian-graphify-rag`) — re-pinned the 8 byte-identity fixtures (config.yaml only, no other drift).
- `memory_tier()` docstring documents the tier-3 mapping.

## Tests

New `test_json_output.py` (preset list, clean JSON-only stdout, tier-3 `rag_endpoint`, `none` empty memory). Full suite green (1517 passed, 3 skipped); `ruff` clean; byte-identity intact.

Closes #510

PR #646, opened 2026-07-08
feat(PI-645): Polish interactive wizard UI (theme, preset table, spinner, summary)

VytCepas:
Closes #645

## What

The interactive `project-init` wizard is the one genuinely human-facing terminal surface. This gives it one visual language and makes it production-ready, `src/`-only (no templates, byte-identity fixtures, or `.claude` mirror touched; `rich` is already a core dep).

- **New `src/project_init/console.py`** — one `rich.Theme` + a single shared `Console`, so the cyan/green/yellow/red vocabulary is named once (`info`/`success`/`warning`/`error`/`muted`/…) instead of retyped across the **~30 lazy `Console()`** instantiations in `__main__.py` (all migrated).
- **Presets → aligned `Table`** with a *Memory* column the old space-padded string list couldn't show.
- **Scaffold → transient spinner** (`console.status` on **stderr**, so it never interleaves with stdout or the `--json` payload).
- **Success screen → one panel** with a key/value grid + a nested file `Tree`, replacing the up-to-four separately-bordered blocks.
- **Unified menu numbering** — delivery/deploy/iac/profile/mcp all route through one `option_line()` helper, killing the `[cyan]1[/cyan].` vs `1. [cyan]name[/cyan]` drift.

## Degrade-to-plain (PI-641)

Every richer device is gated on `is_interactive()` (TTY detection). Piped/redirected/captured runs get plain, unwrapped text — no box-drawing or escape sequences leaking into a transcript, and the `--json` path is byte-identical (early return, untouched).

## Tests

- New `tests/unit/test_console.py` covers `option_line`, the preset table + its plain fallback, missing-memory handling, and the spinner no-op.
- Full suite green: **2000 passed, 8 skipped**.

## Not in scope

Scaffolded-project bash/hook output (separate, higher-risk pass — the Python wizard does not ship into targets). No new dependency, no new CLI flag (ADR-023 partition test unaffected).

### DIFF
--- a/src/project_init/__main__.py
+++ b/src/project_init/__main__.py
@@ -469,7 +469,11 @@ def _choose_preset_interactive(presets: list[dict]) -> dict:
         )
     )
     default_idx = _default_preset_index(presets)
-    render_presets(presets, default_idx)
+    # Resolve each preset's memory stack through extends inheritance so the
+    # table's Memory column is accurate for inheriting presets like `governed`
+    # (raw vars would show "—"; PI-645 review, mirrors #511).
+    memory_by_name = {row["name"]: row["memory_stack"] for row in _presets_payload(presets)}
+    render_presets(presets, default_idx, memory_by_name)
     console.print()
 
     choice = _prompt_menu_index("Choose a preset", len(presets), default=default_idx)
--- a/src/project_init/__main__.py
+++ b/src/project_init/__main__.py
@@ -10,6 +10,13 @@ from datetime import date
 from pathlib import Path
 
 from project_init import __plugin_version__, __repo_url__, __version__
+from project_init.console import (
+    console,
+    is_interactive,
+    option_line,
+    render_presets,
+    scaffolding,
+)
 from project_init.mcps import (
     MCP_CATALOG,
     PLAYWRIGHT_MCP,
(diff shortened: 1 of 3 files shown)


VytCepas:
- `console.py:113` spinner gate — **Fixed.** `scaffolding()` now gates on `is_interactive()` (stdout TTY) **and** `_status_console.is_terminal` (stderr TTY), so 'interactive' is defined consistently with every other richer device. (It never actually leaked — both streams were checked independently — but the framing was inconsistent, so this makes it unambiguous.)
- `test_console.py:25` env-dependent test — **Fixed.** Rewrote to monkeypatch `pc.console` to a non-TTY `io.StringIO()` and assert `is_interactive()` is False deterministically; no longer depends on pytest capture being on (`-s` safe).
- `test_console.py:67` stricter assertion — **Fixed.** Now asserts `captured.err == ""` for the strict 'emit nothing off-TTY' guarantee.

VytCepas:
**Review response (Codex P2 — inherited preset memory):**
- `console.py:96` — **Fixed.** Good catch; this reintroduced the same class of bug as #511. The Memory column read raw per-preset `vars.memory_stack`, so `governed` (which only declares `extends = "obsidian-only"`) showed `—` instead of `obsidian-only`. Now `_choose_preset_interactive` resolves the stack through `_presets_payload` (which applies `load_preset` inheritance) and passes the resolved map into `render_presets`; the presentation layer no longer reads raw vars. Verified `governed` → `obsidian-only` in the rendered table, and added `test_render_presets_shows_resolved_inherited_memory`.

The three earlier Copilot comments were already addressed in the previous commit (consistent spinner gate + deterministic tests). Full suite: 2001 passed, 8 skipped.

PR #648, opened 2026-07-08
feat(PI-647): Propagate token-efficiency conventions to templates + token_efficiency skill

VytCepas:
Closes #647

WS2+WS3 of epic #641 — scaffolded projects get token-frugal working conventions:

- **`test-quick` recipe** in every language justfile block (python `pytest -x -q --tb=short`, node `bun test --bail`, go `-failfast`, rust `-q`) with the same day-one guards as `test`.
- **Token-efficiency rule** in AGENTS.md "Key rules": filter before ingest, line-ranged reads, delegate broad sweeps to the shipped `explore` subagent (with the 4×-total-cost caveat), instruction-file line budgets.
- **New default-on `token_efficiency` skill** (plugin payload + `--no-plugin` fallback per ADR-010; synced to codex/antigravity/amp/junie surfaces): full playbook — input-side command/read/search discipline, output-side specific budgets, env-knob reference (documentation, not overrides). Registered in INDEX/README/project-init.md tables.

Byte-identity: only the new SKILL.md path added to `_ADDED_SINCE_BASELINE` (both modules); AGENTS.md/justfile/INDEX.md already excluded.

Quality guardrail: guidance-only — no guard, hook, or CI behavior changes. `just lint` + full `just test` green (2009 passed).


VytCepas:
**Review response:**
- [Make Python test-quick self-contained]: Fixing — correct catch; `uv run --with pytest pytest -x -q --tb=short` now matches the self-contained pattern of `test`/`test-cov` (xdist intentionally omitted: fail-fast iteration is serial by design). Contract test still asserts the recipe body.

VytCepas:
**Review response:**
- [MEMORY.md may not exist under memory_stack=none, 4 surface copies]: Fixing — the indexes line now says "each where present (not every scaffold ships every index)", fixed once in the source skill (`templates/fallback/.../token_efficiency/SKILL.md`) and re-synced to all surface copies via `tools/sync_plugin.py` (the four flagged files are derived copies of that one source).

PR #650, opened 2026-07-08
feat(PI-649): Deduplicate and trim the workflow_state_reminder injection

VytCepas:
Closes #649

WS1 of epic #641 — the biggest measured token drain (~470 tokens re-injected 10–20×/session).

- **Session-scoped dedup (ADR-028):** sentinel file keyed on sanitized stdin `session_id` + project-dir hash. First trigger injects a **trimmed** rules block (~470 → ~200 tokens: every banned-command → wrapper mapping kept, tutorial prose deferred to the `github_workflow` skill). The sentinel stores a 16-char hash of the DAG state, so repeat triggers are **fully silent unless the state changed** — and inject only the dynamic block when it did.
- **Fail-open:** missing `session_id`, unwritable tmp, or any sentinel I/O error → previous full-injection behavior. Path-traversal in `session_id` is sanitized (tested).
- **Three hook copies in sync:** `.agents/hooks/` (+ `.claude` mirror via sync-claude), `templates/lifecycle_fallback/`, `project-init-lifecycle` plugin (sync_plugin).
- **Enforcement untouched:** the reminder is advisory UX; `github_command_guard.sh` → `dag_workflow.py guard` blocking is unchanged.
- 7 behavioral tests (`tests/integration/test_workflow_state_reminder.py`).

**Also fixes a latent test-infra bug discovered en route:** `TestPrePushLifecycleGate` ran the scaffolded pre-push from pytest's cwd; with `just` installed and a **clean** worktree its `just ci` gate re-ran this repo's suite recursively (forever). Masked locally by dirty trees and in CI by `just` being absent. Fixed with a stub `just` on PATH so the gate deterministically fail-opens.

`just lint` green; full suite 2027 passed / 8 skipped in 10s.

### DIFF
--- a/.agents/hooks/workflow_state_reminder.sh
+++ b/.agents/hooks/workflow_state_reminder.sh
@@ -1,7 +1,13 @@
 #!/usr/bin/env bash
-# workflow_state_reminder.sh — inject the full lifecycle rules when a prompt
+# workflow_state_reminder.sh — inject the lifecycle rules when a prompt
 # mentions GitHub workflow actions, plus the current DAG state if available.
 # UserPromptSubmit hook. Receives prompt JSON on stdin.
+#
+# Token-efficiency (PI-649, ADR-028): injected context persists in the
+# transcript and is re-sent every turn, so the static rules block is injected
+# ONCE per session (sentinel file keyed on the stdin session_id); later
+# triggers get only the dynamic DAG state. Fail-open: no session_id or an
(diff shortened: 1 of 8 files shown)


VytCepas:
**Review response:**
- [strengthen path-traversal assertion]: Fixing — the test now asserts exactly one `pi_wsr_*` sentinel was created **inside** `tmp_path` and that its name ends with the sanitized id (`_etcpasswd`), replacing the brittle `../../etc` existence check.

PR #652, opened 2026-07-08
feat(PI-651): Cap failure-path hook output at 40 lines

VytCepas:
Closes #651

WS4 of epic #641. `post_edit_lint.sh` (PostToolUse `additionalContext`) and `pre_commit_gate.sh` (PreToolUse deny reason) piped raw linter output into context on failure — text that persists in the transcript and is re-sent every turn.

- Injected error text is now capped at **40 lines** inside each hook's Python emitter, with a trailer: "… output truncated (N lines total) — run `just lint` for the full report."
- **Decisions unchanged**: a failing gate still denies; only the reason/context text is capped. The first 40 lines carry the actionable findings.
- Plugin copies re-synced via `tools/sync_plugin.py` (this repo itself doesn't ship these two hooks).
- 3 behavioral tests (`tests/integration/test_hook_output_caps.py`): oversized output capped with trailer (both hooks), small output untouched, deny decision preserved — exercised offline via a bare non-uv temp repo with the venv's ruff.

`just lint` green; full suite 2026 passed / 8 skipped.

### DIFF
--- a/plugins/project-init-workflow/hooks/post_edit_lint.sh
+++ b/plugins/project-init-workflow/hooks/post_edit_lint.sh
@@ -109,7 +109,7 @@ file, errors = sys.argv[1], sys.argv[2]
 lines = errors.splitlines()
 if len(lines) > 40:
     errors = '\n'.join(lines[:40]) + (
-        f'\n… output truncated ({len(lines)} lines total) — re-run the linter on the file for the full report.'
+        f'\n… output truncated ({len(lines)} lines total) — run \`just lint\` for the full report.'
     )
 ctx = f'Lint errors in {file} (after auto-fix attempt) — please fix before continuing:\n{errors}'
 print(json.dumps({'hookSpecificOutput': {'hookEventName': 'PostToolUse', 'additionalContext': ctx}}))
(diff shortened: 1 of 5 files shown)


VytCepas:
**Review response:**
- [trailer should say `just lint`, both copies]: Fixing — `post_edit_lint.sh` trailer now matches `pre_commit_gate.sh` ("run `just lint` for the full report"), fixed in the template source and re-synced to the plugin copy via `tools/sync_plugin.py`.
- [assert the trailer guidance in the cap test]: Fixing — both cap tests now assert `just lint` appears in the truncation trailer.

VytCepas:
**Review response:**
- [argv limit bypasses the cap and the deny on huge reports]: Fixing — excellent catch, and it was a pre-existing gate bypass (a >128KB report made the emitter exec fail with E2BIG, so `pre_commit_gate` never emitted its deny). `$ERRORS` now travels via stdin in both hooks (template + plugin copies re-synced). New regression test stages 3000 ruff errors (~165KB report, above Linux's per-argv cap) and asserts the capped deny is still emitted.

PR #656, opened 2026-07-08
feat(PI-655): Attribute tool_result context volume per tool in usage_report

VytCepas:
Closes #655

WS6 of epic #641 — the measurement piece that lets WS1/WS4/WS5 be verified against real sessions.

- `parse_transcript` now sums `tool_result` content size (chars) per tool name during the existing single pass, via the existing `tool_use_id → name` map. Handles both string and typed-block content forms. **Aggregate sizes only — no content kept** (existing privacy rule).
- New "Top context contributors" section in both text (`chars` + ≈tokens at 4 chars/token) and HTML (bar chart, top 10) renders, under the Cost bucket (`context_contributors`).
- Per-hook attribution deliberately out of scope: `usage.jsonl` already counts hook firings and payloads are static.
- Tests: fixture transcript extended with a typed-block tool_result; asserts exact per-tool char attribution and section presence in both renders.

`just lint` green; full suite 2023 passed / 8 skipped.

### DIFF
--- a/templates/observability/dot_agents/observability/usage_report.py
+++ b/templates/observability/dot_agents/observability/usage_report.py
@@ -160,10 +160,14 @@ def _content_chars(content: object) -> int:
     if isinstance(content, str):
         return len(content)
     if isinstance(content, list):
+        # A malformed block (non-string "text") counts 0 instead of aborting
+        # the whole report — same resilience posture as _safe_int.
         return sum(
-            len(b.get("text") or "")
+            len(text)
             for b in content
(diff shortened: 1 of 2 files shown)


VytCepas:
**Review response:**
- [`_content_chars` TypeError on non-string `text`]: Fixing — a malformed typed-block now counts 0 instead of aborting the report (same resilience posture as `_safe_int`); the fixture transcript gained a `{"type": "text", "text": 123}` block and the per-tool assertion proves it's ignored.

PR #660, opened 2026-07-09
chore(PI-659): Dedupe project-init.md format table and compaction list

VytCepas:
Closes #659

WS11 of epic #641 (conservative scope). Two verbatim duplications removed from `project-init.md.tmpl` — three drifting copies of the same rules become one source each:

- **Commit-and-PR format table** → pointer to the AGENTS.md quick-ref (single source); types list and the nojira flow stay inline (not duplicated elsewhere at that detail).
- **Compaction preserve-list** → pointer to CLAUDE.md (compaction is a Claude-only mechanism; the canonical list lives there).

**Scope guardrail:** the rest of the GitHub-tracking section deliberately STAYS — it is the lifecycle documentation for skill-less surfaces (Cursor, VS Code, Ollama agents); removing it would be a quality loss.

Tests: `test_project_init_md_dedupe.py` (pointer present, duplicates gone, canonical copies intact, links resolve). `just lint` green; full suite 2035 passed / 8 skipped.

### DIFF
--- a/templates/base/dot_agents/project-init.md.tmpl
+++ b/templates/base/dot_agents/project-init.md.tmpl
@@ -108,9 +108,9 @@ This project is **single-trunk**: feature PRs target `main` and merge by squash.
 ### Commit and PR format
 
 Commit messages, PR titles, and branch names follow the quick-ref in
-[AGENTS.md](../AGENTS.md) (single source: `type(KEY-N): description`, branch
-`<type>/<KEY>-<n>-<slug>`, body includes `Closes #<n>`). Commit messages use
-the same format as PR titles. Types: `feat` · `fix` · `chore` · `docs` · `test`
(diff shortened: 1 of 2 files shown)


VytCepas:
**Review response:**
- [pointer still restates formats, with a `Closes #<n>` vs `Closes #N` inconsistency]: Fixing — the parenthetical restatement is gone; the paragraph now carries only the pointer, the commit-equals-PR-title note, and the types list. AGENTS.md is the sole copy of the format strings.

VytCepas:
**Review response:**
- [link test doesn't check actual hrefs]: Fixing — the test now extracts the rendered markdown hrefs ending in AGENTS.md/CLAUDE.md, asserts both pointers exist as real links, and resolves each relative to the file's own location (a plain `AGENTS.md` href would now fail).

PR #664, opened 2026-07-09
feat(PI-663): checkpoint skill — checkpoint-and-clear session handoff

VytCepas:
Closes #663

WS9 of epic #641 — attacks the session-longevity multiplier: every turn re-sends the whole conversation, so a long session pays 100k+ history tokens per exchange. The `checkpoint` skill writes a structured handoff (task, verified work, decisions, next steps, branch/files, test state) to `.agents/tmp/checkpoint.md`; the user `/clear`s and resumes from ~500 tokens with nothing load-bearing summarized away.

**Hard requirements from the epic review, both enforced by test:**
- `.agents/tmp/` is **gitignored by the scaffold** (`dot_gitignore.tmpl`) — session state is never committable by accident (`git check-ignore` test).
- The skill **offers deletion after resume** (a stale checkpoint silently misleads the next resume).

Also: prescribed handoff structure (~60-line cap; long content belongs in commits/ADRs/memory), resume protocol (decisions are settled; verify branch + test state before continuing), and a comparison table vs `/compact` and `session_summary`. Ships via the ADR-010 plugin/fallback split + all agent surfaces; INDEX/README/project-init rows added.

Byte-identity: new SKILL.md → exclusions; the `.gitignore` edit re-pinned **only that key** across all 8 fixtures (scaffold-as-fixtures-do recompute).

`just lint` green; full suite green.


VytCepas:
**Review response:**
- [old scaffolds get the skill via plugin update but lack the ignore entry]: Fixing — excellent catch. The skill now runs `git check-ignore -q .agents/tmp/checkpoint.md` before writing and appends the `.agents/tmp/` entry to .gitignore itself when missing; asserted in the content test.

PR #670, opened 2026-07-09
feat(PI-666): CI_RUNS_ON self-hosted-runner escape hatch for scaffolded ci.yml

VytCepas:
Closes #666 (PR 1 of 2 per the implementation plan in the issue; PR 2 = the `local_ci` detection/suggestion skill follows separately)

When a **private** repo exhausts its GitHub Actions minutes, every job fails at start with the billing message and there's no way onto a self-hosted runner short of hand-editing every workflow. Self-hosted runners bill **zero** minutes and keep checks reporting, so branch protection stays intact.

- **`ci.yml.tmpl`**: 11 compute jobs now read `runs-on: ${{ vars.CI_RUNS_ON || 'ubuntu-24.04' }}` — one repo variable flips CI to a self-hosted runner, unset restores GitHub-hosted. **Pinned with in-file reasons:** `scorecard` (OSSF action requires GitHub-hosted) and `ci-gate` (seconds-long aggregator).
- **Security invariant, tested in both directions** (`test_ci_runs_on.py`): every unpinned ci.yml job uses the variable, AND the secret/PAT-bearing lifecycle workflows (`board-automation`, `validate-pr`, `issue-validation`, `project-init-upgrade`) provably never contain `CI_RUNS_ON` — tokens never materialize on a user's machine (per the security review noted in the issue).
- **`just ci-local-on` / `ci-local-off`** toggles (thin `gh variable` wrappers) in every language block.
- **New scaffolded guide** `.agents/docs/guides/self-hosted-ci-runner.md(.tmpl)`: diagnosis commands (`gh run list` startup_failure, billing API), runner registration with **`--ephemeral` recommended**, the trust model (private + trusted collaborators only; never public forks), the don't-set-the-variable-without-a-live-runner warning, and honest alternatives.
- **`--lifecycle none` correctness**: workflow name-drops in prose are gated behind `{{#if lifecycle}}` (caught by the existing `test_lifecycle_none` dangling-reference contract during development).

Byte-identity: ci.yml/justfile already excluded; guide added to both exclusion sets.

`just lint` green; full serial suite 2057 passed / 5 skipped.

### DIFF
--- a/templates/base/dot_agents/docs/guides/self-hosted-ci-runner.md.tmpl
+++ b/templates/base/dot_agents/docs/guides/self-hosted-ci-runner.md.tmpl
@@ -1,0 +1,78 @@
+# Self-hosted CI runner — the Actions-minutes escape hatch (PI-666)
+
+When a **private** repo exhausts its GitHub Actions minutes (or hits a
+spending-limit/payment problem), every workflow job fails at start:
+*"The job was not started because recent account payments have failed or your
+spending limit needs to be increased."* Self-hosted runners bill **zero**
+minutes on private repos, and checks keep reporting to GitHub — branch
+protection and the merge gate stay intact.
+
+The scaffolded `ci.yml` reads `vars.CI_RUNS_ON` on its compute jobs; one repo
+variable routes them to your machine. Nothing else changes.
+
+## Diagnose first
+
+```bash
+gh run list --limit 5 --json conclusion,displayTitle   # billing lockout shows startup_failure
+gh api /users/{username}/settings/billing/actions      # included vs used minutes (org repos: /orgs/{org}/...)
+```
+
+Meanwhile, verification never stops: `just ci` runs the exact same gate
+locally (the recipes are what CI runs), and the pre-push git hook already
+enforces it.
+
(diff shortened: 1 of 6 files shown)


VytCepas:
**Review response:**
- [P1: required CI gate pinned defeats the escape hatch]: Fixing — completely right; I'd optimized the gate for cost when availability is what matters: pinned, the REQUIRED check can't start during the lockout and the merge stays blocked. `ci-gate` now follows `CI_RUNS_ON` (in-file comment explains why), and the test asserts it uses the variable while `scorecard` (schedule-gated, never blocks a PR, OSSF-hosted requirement) remains the only pin.

VytCepas:
**Review response:**
- [line-spanning conditional could splice a non-comment line]: Fixing in spirit — the reported breakage doesn't reproduce (verified by rendering with lifecycle off: the engine splices the span cleanly and the line stays commented), but the inline-across-lines form was untested and fragile against engine changes. Restructured to the whole-line conditional pattern and added `test_header_renders_as_valid_comments_without_lifecycle`, which renders a `--lifecycle none` scaffold and asserts every header line is a comment.

PR #667, opened 2026-07-09
feat(PI-665): diagram skill — collaborative Mermaid-first diagramming

VytCepas:
Closes #665

A `/diagram` skill for iteratively drawing architecture, code, data-model, and idea diagrams with the agent — **source files first** (diffable, regenerable), renders as previews.

- **Notation-by-type table**: flowchart/sequence/state/ER/class/mindmap/gantt in Mermaid (`layout: elk` for dense graphs); honest escalations — DOT source (renders only if `dot` exists), Excalidraw JSON for spatial idea sketches (no local render, says so).
- **Ground before drawing**: code diagrams derive from CODE_MAP/imports/structure — a box not backed by a nameable file/module doesn't go on the diagram; idea mode interviews for entities/relations first.
- **Iteration loop**: source under `docs/diagrams/` (or the vault's `design/` — its existing convention), side-panel Mermaid preview on Claude Code / `bunx @mermaid-js/mermaid-cli` SVG elsewhere, ONE targeted question per round, stable node IDs so diffs stay reviewable.
- **Quality rules**: ≤25 nodes per view (overview + drill-downs), subgraphs for boundaries, labeled edges, no color-only meaning.
- **Embedding**: GitHub renders fences natively; mkdocs superfences snippet offered on demand (not a template default — keeps this PR skill-only).

**Live-smoked before push**: rendered a real diagram of project-init's scaffold pipeline via `bunx @mermaid-js/mermaid-cli` (cached Puppeteer Chrome — no downloads beyond the JS package) and previewed the `.mmd` in the Claude side panel. Also installed user-level and invoked in-session.

Ships via ADR-010 (plugin + fallback + codex/antigravity/amp/junie). Byte-identity: SKILL.md added to exclusions in both modules; no fixture re-pins. Tests: `test_diagram_skill.py` (frontmatter, method markers, three discoverability tables, <500-line cap).

`just lint` green; full serial suite (CI-equivalent) 2049 passed / 8 skipped.

PR #672, opened 2026-07-09
feat(PI-671): local_ci skill + monitor_pr billing-lockout hint

VytCepas:
Closes #671 — PR 2 of the plan in #666 (PR 1 = #670, merged).

The escape hatch (#670) only helps if the user finds it. This PR makes project-init notice the lockout and suggest it:

- **`local_ci` skill** (ADR-010 split + all surfaces): (1) diagnose — `gh run list` `startup_failure` signature + billing API usage check, with an explicit "if runs fail for any other reason this skill does not apply"; (2) immediate unblock — `just ci` locally IS the CI gate (recipes identical by design) and the pre-push hook already enforces it; (3) durable fix — runner guide + `just ci-local-on`, with the don't-set-before-the-runner-is-online warning; (4) honest alternatives; (5) safety rails (PAT workflows stay hosted, no self-attested checks, trusted-collaborators-only).
- **`monitor_pr.sh`** (template + repo copy): the fail-closed CI-timeout branch now checks recent runs for `startup_failure` and prints a 3-line pointer at the skill. Healthy runs: zero behavior change (asserted).
- Workflow plugin **0.4.0** (payload change, ADR-010).
- Tests: skill content + discoverability tables; monitor hint via stubbed `gh` — both the firing path (startup_failure → hint) and the non-firing path (healthy → no hint, pre-existing timeout guidance intact).

`just lint` green; full serial suite 2062 passed / 5 skipped.

### DIFF
--- a/.agents/scripts/monitor_pr.sh
+++ b/.agents/scripts/monitor_pr.sh
@@ -189,6 +189,13 @@ for c in json.load(sys.stdin):
     if c.get('name') != 'review/decision' and c.get('bucket') == 'pending':
         print('  -', c.get('name'))" 2>/dev/null || true
     echo "Re-run once the check registers, or investigate why it never started."
+    # PI-671: a billing/minutes lockout means checks NEVER register — point
+    # at the escape hatch instead of leaving the user to rediscover it.
+    if gh run list --limit 5 --json conclusion --jq '.[].conclusion' 2>/dev/null | grep -q startup_failure; then
+      echo "Note: recent runs show startup_failure — GitHub Actions may be out of"
+      echo "minutes or billing-locked. Load the local_ci skill (self-hosted runner"
+      echo "escape hatch: just ci-local-on)."
+    fi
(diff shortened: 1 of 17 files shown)


VytCepas:
**Review response:**
- [legacy billing endpoint]: Fixing — verified against current GitHub docs: the enhanced-billing endpoints are `GET /users/{username}/settings/billing/usage` / `GET /organizations/{org}/settings/billing/usage` (filter `usageItems` for `product == "actions"`). The skill now leads with those and keeps the legacy `/settings/billing/actions` as a one-line fallback for accounts not yet migrated. Fixed once in the source skill and re-synced to all surface copies.

VytCepas:
**Review response:**
- [ineffective case-mixed assertion]: Fixing — same bug class I fixed in PR #667's review; now a single exact-case assertion on the full safety-rail phrase including `CI_RUNS_ON`.

PR #675, opened 2026-07-09
fix(PI-673): semgrep via uvx — pip install fails on self-hosted runners

VytCepas:
Closes #673

Downstream discovery (zarija #115 / PR #116) after the CI_RUNS_ON escape hatch shipped: the scaffolded semgrep job ran `pip install semgrep`, which fails on self-hosted runners (externally-managed environment) — exactly where the toggle now routes it. Also violated the scaffold's own "uv/bun only, no pip" principle.

- `setup-uv` step (same pinned action the other jobs use) + `SEMGREP="uvx semgrep@1.168.0"` — ephemeral pinned env, identical behavior on hosted and self-hosted.
- Test: rendered ci.yml has no `pip install` on executable lines (the explanatory comment legitimately names the failure mode) and uses `uvx semgrep@`.

`just lint` green; full serial suite green (2063 passed equivalent — one new test).


VytCepas:
**Review response:**
- [unpinned setup-uv]: Fixing — pinned to `version: "0.11.7"`, matching every other setup-uv step in the workflow.

PR #676, opened 2026-07-09
fix(PI-674): PI_CI_TIMEOUT override for the monitor CI wait

VytCepas:
Closes #674

Downstream discovery (zarija recap after #666): a single self-hosted runner serializes CI jobs (~15–20 min), exceeding `monitor_pr.sh`'s hardcoded 900s wait — the lifecycle then fails closed every time and merges go manual, defeating the wrapper scripts in exactly the setup the escape hatch recommends.

- Both copies (template lifecycle + repo) read `CI_TIMEOUT="${PI_CI_TIMEOUT:-900}"` — default unchanged, override documented in-file.
- Runner guide: single-runner note (second runner restores parallelism; `PI_CI_TIMEOUT=1800` keeps the merge flow alive), gated behind `{{#if lifecycle}}` — the dangling-reference contract caught the ungated first draft, same as on #670.
- Test (`test_monitor_ci_timeout.py`, stubbed gh + no-op sleep): `PI_CI_TIMEOUT=20` → the timeout message reports 20s, still fails closed.

`just lint` green; full serial suite green after the gating fix (the one failure was this contract).


VytCepas:
**Review response:**
- [unvalidated PI_CI_TIMEOUT crashes the -ge comparison]: Fixing — both copies now validate with the same case-pattern the script uses for --review-cycle: empty/non-numeric/zero → clear stderr message + exit 2 (fail closed). New test asserts the invalid path.

VytCepas:
**Review response:**
- ["00" passes the literal-0 pattern]: Fixing — the case pattern now only rejects empty/non-digits; a numeric `-eq 0` check catches 0/00/000. Test loops over twenty/0/00.

PR #693, opened 2026-07-09
fix(PI-691): create_issue.sh: host-aware --parent URLs and --agent-ready validation

VytCepas:
Closes #691

### DIFF
--- a/.agents/scripts/create_issue.sh
+++ b/.agents/scripts/create_issue.sh
@@ -50,6 +50,7 @@ VALID_SCALES="epic task"
 VALID_PRIORITIES="high medium low"
 VALID_SIZES="XS S M L XL"
 VALID_CONFIDENCES="high medium low unknown"
+VALID_AGENT_READY="Yes No"
 
 usage() {
   cat <<'EOF'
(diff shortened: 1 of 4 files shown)


VytCepas:
**Review response:**
- [Codex P2 — parent URL host discarded]: **Fixed** (76cc8e5). Correct and a genuine correctness hole in my first pass: `parse_parent` now sources `gh_host.sh`, compares the URL host against the active host, and fails loudly on mismatch with an actionable message (`PROJECT_INIT_HOST`/`GH_HOST` override). Deterministic offline test added (`PROJECT_INIT_HOST` pins the host, no network).
- [Copilot ×3 — `contains_word` grep-option injection]: **Fixed** (76cc8e5) at the helper, not per-callsite: values are matched `grep -qFx --` (fixed-string, whole-line, option-safe), which hardens every validation flag, not just `--agent-ready`. The three copies are one finding — template is the source; `.agents/` and `.claude/` copies propagate via `just sync-agents` + `just sync-claude` (enforced by `test_agents_template_sync.py`).

PR #697, opened 2026-07-09
feat(PI-696): PostToolUse tool-output compressor for oversized git-diff-class results (prototype)

VytCepas:
Closes #696


VytCepas:

PR #699, opened 2026-07-09
fix(PI-698): Fresh scaffold fails its own ruff gate: compressor hook PTH violations + pre-edit guard S-rules

VytCepas:
Closes #698


VytCepas:

PR #701, opened 2026-07-09
chore(PI-700): Reduce scaffold turn-tax: actionable guard denials, gate batching, benchmark-derived token guidance, profile routing

VytCepas:
Closes #700


VytCepas:

PR #702, opened 2026-07-09
fix(PI-631): fix: start_issue.sh derives the project key from the current directory name, not config

VytCepas:
Closes #631


VytCepas:
**Review response:**
- Bare-worktree key derivation (P2): Fixing — `_repo_root_name` now branches on the common dir's basename: `.git` → parent dir name (normal layout), `*.git` → the bare dir's own name minus the suffix, anything else (custom `GIT_DIR`) → current-worktree fallback as suggested. Added a bare-repo worktree behavioral test (`test_helper_handles_bare_repo_worktrees`).

PR #703, opened 2026-07-09
fix(PI-668): Upgrade round-trip tests flake under pytest-xdist full-suite runs

VytCepas:
Closes #668

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,10 @@ logs/
 !.agents/docs
 .agents/docs/*
 !.agents/docs/CODE_MAP.md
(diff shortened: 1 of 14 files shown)


VytCepas:
**Review response:**
- Patch aliases imported before fixtures run (P2): Fixing — replaced the two named module patches with a `sys.modules` sweep that re-points every loaded module whose `_TEMPLATES_DIR` is the original object (identity match), restoring symmetrically on teardown. Covers `test_docs_renovate`/`test_justfile_combinations` and any future alias without enumeration; modules imported after the fixture from-import the already-patched global. Re-validated: full xdist suite green with the template-toggler running.

PR #705, opened 2026-07-09
fix(PI-632): fix: finish_pr.sh / monitor_pr.sh merge step races the last CI check

VytCepas:
Closes #632

### DIFF
--- a/.agents/scripts/monitor_pr.sh
+++ b/.agents/scripts/monitor_pr.sh
@@ -142,8 +142,40 @@ _run_gh() {
   return "$status"
 }
 
+# The PR being MERGED is success regardless of how the last attempt exited —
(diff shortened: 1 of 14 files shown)


VytCepas:
**Review response:**
- `_pr_is_merged` robustness: Fixing — added `GH_PROMPT_DISABLED=1` and `|| true` inside the substitution so any gh failure reads as 'not merged' (fail-closed to retry, never abort). Note: under `set -e` a failed substitution inside `[ ... ]` doesn't abort the script, but the explicit guard documents the intent and covers `inherit_errexit` setups.
- Contract test line-index tracking: Fixing — the helper span is now bounded by line index (body ends at the first bare `}`), so a text-identical `pr merge` line elsewhere can't be mistaken for the helper's own attempts.

PR #707, opened 2026-07-09
fix(PI-678): finish_pr/monitor_pr: delete the merged local branch after a successful merge

VytCepas:
Closes #678

### DIFF
--- a/.agents/scripts/monitor_pr.sh
+++ b/.agents/scripts/monitor_pr.sh
@@ -151,6 +151,10 @@ _run_gh() {
 # skips silently when the branch is absent and loudly when it diverged.
 _cleanup_local_branch() {
   local head_info head_ref head_oid base local_oid current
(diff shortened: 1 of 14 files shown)


VytCepas:
**Review response:**
- Merge-queue enqueue vs merge (P2): Fixing — `_cleanup_local_branch` now starts with `_pr_is_merged || return 0`, so cleanup only runs once the server reports the PR MERGED; an enqueued-but-open PR keeps its local branch (new test `test_enqueued_but_unmerged_pr_keeps_branch`).
- `-d` before `-D`: Fixing — safe delete attempted first; force only when squash ancestry makes `-d` refuse, backed by the SHA-equality guard.
- Silent dirty skip: Fixing — the dirty-worktree case now returns silently per #678's acceptance criteria. Not applying the global dirty check, though: dirtiness only matters when cleanup would switch branches (the on-branch case); deleting a non-checked-out branch never touches the worktree.

Issue #251, opened 2026-06-17
Hybrid enforcement: classify guardrails enforced-vs-overridable and make the hard layer bind by default (org profile)


VytCepas:
## Summary

Realize hybrid enforcement. Today every Claude/git hook is soft/bypassable and branch protection is opt-in, so the "non-negotiable" half is cosmetic until someone runs `setup_github.sh --protect`. Deliver:
1. A documented taxonomy classifying each guardrail HARD (server-side: CI required checks, branch protection / org rulesets) vs SOFT (overridable: Claude hooks, local git hooks).
2. In the **org profile**, make the HARD layer bind by default — bootstrap branch protection / rulesets and required checks at scaffold (or a one-command setup), so enforced items can't be locally bypassed.
3. A CI job that verifies the required guardrails/workflows are present (tamper/drift detection), since client-side locking is not reliable (ADR-007).

Honest constraint: do NOT attempt to lock `settings.json` client-side — that breaks the plugin model. Enforcement lives server-side; the profile just makes it on-by-default for orgs.

## Metadata
- Type: feature
- Priority: high
- Size: L
- Area: templates (ci.yml, validate-pr.yml, setup_github.sh, rulesets) + config + docs
- Agent ready: No
- Confidence: medium

## References
- ADR-007 (enforcement layers; CI is the real boundary)
- #145 (governance bootstrap: CODEOWNERS, SECURITY, branch protection)
- ADR-013

## Dependencies
- ADR-013; pairs with the profile switch

## Acceptance criteria
- [ ] A taxonomy doc marks each guardrail enforced-vs-overridable with rationale.
- [ ] org profile enables the HARD layer by default (branch protection / required checks / rulesets bootstrap).
- [ ] A CI check verifies required guardrails/workflows exist and flags removal.
- [ ] SOFT guardrails remain overridable; no client-side settings.json locking is attempted.

## Definition of Ready
- [ ] Agree the exact enforced (non-negotiable) set vs the overridable set.

## Definition of Done
- [ ] Taxonomy documented; org-profile hard defaults implemented; guardrail-presence CI check added; tests cover profile rendering.

---

## Grill+Codex revision (2026-06-18)

- **Split the taxonomy by primitive** — branch protection, **branch/tag rulesets**, **push rulesets**, required workflows — each with a **feature probe** for availability per tier (corrects the "rulesets = Enterprise-only" oversimplification).
[...]

PR #709, opened 2026-07-09
chore(PI-708): Reconcile monitor_pr.sh divergence: sync root copy from the template

VytCepas:
Closes #708


VytCepas:
**Review response:**
- Admin-merge refusal message line wrapping: Not applying — the text is pre-existing template content (shipped with the org-profile guard, #251); this PR only syncs it into the root copy verbatim, which is the whole point of the reconciliation. Rewording it here would re-diverge nothing but would touch the template and re-pin 8 byte-identity fixtures for a cosmetic echo-wrapping change; happy to fold it into a future template content pass.

PR #712, opened 2026-07-09
fix(PI-706): gate the CI verdict on the API head matching the pushed tip

VytCepas:
## Problem

`gh pr checks` reports the rollup for whatever commit the API believes is the PR head. Right after a push, the API can still serve the **previous** `headRefOid` (replication lag). That commit's checks are already settled, so the CI wait loop in `monitor_pr.sh` broke on its first poll and judged the wrong commit:

- a **red** predecessor read as "CI failed" — observed on #705, where the new run went green a minute later;
- a **green** predecessor would have merged a commit whose CI never ran.

The existing empty-list guard only covers "no checks yet"; it cannot distinguish stale-but-settled from done.

## Fix

Before any check result is trusted, wait until the API's `headRefOid` agrees with the branch tip that `git ls-remote` reports. `ls-remote` answers from git's endpoint rather than the REST/GraphQL API, so it sees the pushed tip immediately.

The gate **fails open** whenever the expected SHA cannot be established — a fork PR, a deleted branch, no `origin` — leaving behavior exactly as before. `PI_HEAD_SYNC_TIMEOUT` (default 120s, `0` disables) bounds the wait and fails closed on expiry.

Shared lifecycle script: edited in `templates/`, synced to `.agents/` and `.claude/` via `just sync-agents` + `just sync-claude` (PI-685).

## Testing

New `tests/integration/test_monitor_head_sync.py` stubs `gh` and `git ls-remote` to cover four cases: stale API head never reaches the check verdict; synced head passes through to it; `PI_HEAD_SYNC_TIMEOUT=0` disables the gate; an invalid timeout fails closed.

Full suite green (2122 passed), `ruff check` clean, shellcheck gate passes.

Closes #706

### DIFF
--- a/.agents/scripts/monitor_pr.sh
+++ b/.agents/scripts/monitor_pr.sh
@@ -301,16 +301,34 @@ _has_review_activity() {
 # predecessor reads as "CI failed" (observed on #705), and a green one would
 # merge a commit whose CI never ran. `git ls-remote` answers from git's endpoint
 # rather than the API, so it sees the pushed tip immediately: wait until the two
-# agree before any check result is trusted. Fails open (returns 0) whenever the
-# expected SHA cannot be established — a fork PR, a deleted branch, no `origin`
-# — leaving behavior exactly as it was before this gate existed.
+# agree before any check result is trusted.
+#
+# Two distinct outcomes when the SHAs don't line up:
+#   * skipped entirely (return 0, unchanged behavior) when no expected SHA can
+#     be established — a cross-repo PR, a deleted branch, no `origin`, or the
+#     gate switched off with PI_HEAD_SYNC_TIMEOUT=0;
+#   * fail closed (exit 1) when an expected SHA IS known and the API never
+#     catches up to it within the timeout, since every check result on hand
+#     may belong to some other commit.
 _wait_for_head_sync() {
-  local head_ref remote_sha api_sha elapsed
+  local head_info head_ref cross_repo remote_sha api_sha elapsed
   if [ "$HEAD_SYNC_TIMEOUT" -eq 0 ]; then
     return 0
   fi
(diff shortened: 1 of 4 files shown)


VytCepas:
**Review response:**

- **Codex — use the PR head repo when resolving the remote SHA (P2)**: Fixing. This is a real bug, not just a robustness nit. For a cross-repo PR, `headRefName` is only the branch name, so `git ls-remote origin refs/heads/$head_ref` resolves the *base* repo's same-named branch (`main`, `feature`). The gate would then compare the fork's head SHA against a SHA from the wrong repository, never converge, and exit 1 at the timeout — the opposite of the fail-open behavior the comment promised. Now the gate reads `isCrossRepository` alongside `headRefName` and returns early when it is true. The fork's remote isn't configured locally, so skipping is the only correct answer; guessing a remote would be worse. Covered by `test_cross_repo_pr_skips_the_gate`.

- **Copilot — "fails open" comment contradicts the fail-closed timeout path**: Fixing. The comment was describing only the skip path. Rewritten to name both outcomes explicitly: *skipped* (return 0, unchanged behavior) when no expected SHA can be established — cross-repo PR, deleted branch, no `origin`, or `PI_HEAD_SYNC_TIMEOUT=0` — and *fail closed* (exit 1) when an expected SHA is known and the API never catches up, since the checks on hand may belong to another commit. Also corrected the `PI_HEAD_SYNC_TIMEOUT` comment, which cited forks as the escape-hatch case; forks are now handled directly.

- **Copilot — `shutil.which("git")` can return `None`**: Fixing. It would have rendered the stub as `exec None "$@"`. Asserted with a message naming the missing binary.

Full suite green, `ruff check` clean, shellcheck gate passes.

PR #713, opened 2026-07-09
fix(PI-628): pin one Python version across mise, mypy, and the CI matrix

VytCepas:
## Problem

A single greenfield Python scaffold emitted **three different answers** to "what Python is this project on":

| File | Value | Where it came from |
|---|---|---|
| `mypy.ini` | `3.11` | `python_floor`, whose default is 3.11 when no `pyproject.toml` exists — the normal greenfield case |
| `mise.toml` | `3.12` | hardcoded in `mise.toml.tmpl` |
| CI matrix | `3.11`–`3.14` | the full `KNOWN` set, since `requires-python` is absent |

So on day one mise provisions 3.12, mypy checks against 3.11 typeshed, and CI tests four versions — and none of them is necessarily the one the user intends.

## Fix

One value feeds all three. `--python-version` (choices `3.11`–`3.14`) is a **mechanical** flag per ADR-023 — an input value like `--name`, not a selectable concern, so it needs no explainer panel; `test_wizard_explanations.py` classifies it via `WIZARD_MECHANICAL_FLAGS`.

**Resolution order:**

1. A declared `requires-python` in `pyproject.toml` is **authoritative** — the scaffolded CI re-derives its matrix from that file on every run, so nothing else can outrank it.
2. `--python-version` supplies the value when nothing declares one.
3. Otherwise `3.11`, the oldest supported CPython.

A `--python-version` that **contradicts** a declared `requires-python` is rejected before the target directory is created, rather than half-applied. (Caught in review — see below.) The wizard asks only on a greenfield Python scaffold; when `requires-python` exists it stays silent, and non-Python scaffolds never see the prompt.

The CI template's no-`requires-python` branch now starts the matrix at the scaffolded floor rather than fanning across the full `KNOWN` set, so a `--python-version 3.13` project tests `[3.13, 3.14]`.

A contract test pins `KNOWN` in `ci.yml.tmpl` byte-equal to `SUPPORTED_PYTHON_VERSIONS` in the scaffolder, so a user can never pin a floor CI won't test.

## Review finding (Codex, P2)

The first revision let `--python-version` win over a declared `requires-python` for `mise.toml` and `mypy.ini` — but the generated CI still entered its `if spec:` branch and derived the matrix from `requires-python`. Reproduced: `requires-python = ">=3.12"` + `--python-version 3.14` rendered both local pins at 3.14 while CI tested `['3.12', '3.13', '3.14']`. mypy would then type-check against 3.14 typeshed and green-light syntax the 3.12 job cannot run.

Making CI follow the rendered floor instead would have traded one incoherence for another — CI testing only 3.14 while the package metadata still advertised 3.12 support. `requires-python` is the project's public contract and project-init does not own `pyproject.toml`, so the conflict now **fails closed** with a message naming both values and the two ways out.

## Verification

Beyond the tests, I scaffolded real projects and executed the rendered CI matrix script in a clean cwd:

- greenfield, no flag → all three read `3.11`, matrix `[3.11, 3.12, 3.13, 3.14]`
- `--python-version 3.13` → all three read `3.13`, matrix `[3.13, 3.14]`
- existing `requires-python = ">=3.12"`, no flag → all three read `3.12`
- existing `requires-python = ">=3.12"` + `--python-version 3.14` → rejected, nothing written
- existing `requires-python = ">=3.12"` + `--python-version 3.12` → accepted
- `--python-version 3.9` → rejected before the target dir is created
- wizard: prompts for Python, silent for Go, silent when `requires-python` exists

Three scripted-wizard tests needed an extra answer because a new prompt was inserted; each is annotated.

Full suite green (2134 passed), `ruff check` clean.

Closes #628

### DIFF
--- a/docs/guides/using-project-init.md
+++ b/docs/guides/using-project-init.md
@@ -54,9 +54,11 @@ The most-used flags:
 
 `--python-version` is the single source for "what Python is this project on": it
 pins `mise.toml`'s toolchain, `mypy.ini`'s typeshed baseline, and the floor of the
-CI test matrix to one value. On a greenfield scaffold the wizard asks; once your
-`pyproject.toml` declares `requires-python`, that becomes the source of truth and
-CI derives the matrix from it instead.
+CI test matrix to one value. On a greenfield scaffold the wizard asks. Once your
+`pyproject.toml` declares `requires-python`, that file is authoritative — CI
+re-derives the matrix from it on every run — so the wizard stops asking, and a
+`--python-version` that contradicts it is rejected rather than half-applied.
+To move an existing project, edit `requires-python` and re-run `project-init upgrade`.
 
 The full surface (delivery/deploy/IaC overlays, governance, observability,
 multi-model, agents, license, profile, …) is documented by `project-init --help`
--- a/docs/guides/using-project-init.md
+++ b/docs/guides/using-project-init.md
@@ -45,11 +45,18 @@ The most-used flags:
 |------|--------|---------|
 | `--preset` | `core`, `auto`, `obsidian-only`, `obsidian-graphify`, `governed` (see `--list-presets`) | (asked interactively) |
 | `--language` | `python`, `node`, `go`, `rust`, `none` | `none` |
+| `--python-version` | `3.11`, `3.12`, `3.13`, `3.14` | `pyproject.toml`'s `requires-python` floor, else `3.11` |
 | `--memory` | `none`, `auto`, `obsidian-only`, `obsidian-graphify`, `obsidian-graphify-rag` | the preset's tier |
 | `--lifecycle` | `github`, `none` | `github` |
 | `--mcps` | `context7`, `context7-http` (comma-separated) | none |
 | `--browser` | flag (Playwright MCP) | off |
 | `--strict` | flag (fail on unrendered placeholders) | off |
+
+`--python-version` is the single source for "what Python is this project on": it
+pins `mise.toml`'s toolchain, `mypy.ini`'s typeshed baseline, and the floor of the
+CI test matrix to one value. On a greenfield scaffold the wizard asks; once your
+`pyproject.toml` declares `requires-python`, that becomes the source of truth and
+CI derives the matrix from it instead.
 
 The full surface (delivery/deploy/IaC overlays, governance, observability,
 multi-model, agents, license, profile, …) is documented by `project-init --help`
--- a/src/project_init/__main__.py
+++ b/src/project_init/__main__.py
@@ -2677,7 +2677,12 @@ def _cli(argv: list[str]) -> int:
     _reject_bare_subcommand_target(args.target, parser)
     target = Path(args.target).resolve()
     # Before the target directory is created (PI-20) and before any prompt.
-    _reject_python_version_without_python(args.python_version, args.language, parser)
+    # --language is optional; a non-interactive run resolves an absent one to
+    # "none" (see _resolve_inputs), so validate against that rather than letting
+    # None read as "unknown" and slip the flag through (PR #713 review). In an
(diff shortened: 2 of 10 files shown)


VytCepas:
**Review response:**

- **Codex — honor --python-version in CI for existing pyprojects (P2)**: Fixing, and thank you — this was a real defect, and it broke the exact "one value feeds all three" contract this PR claims. I reproduced it: `requires-python = ">=3.12"` + `--python-version 3.14` rendered `mise.toml` and `mypy.ini` at 3.14 while the generated CI still entered its `if spec:` branch and tested `['3.12', '3.13', '3.14']`. The consequence is worse than cosmetic drift: mypy would type-check against 3.14 typeshed and green-light syntax that the 3.12 job cannot run.

  I did **not** take the suggested direction of making CI follow the rendered `{{python_floor}}`. `requires-python` is the project's public support contract, and project-init deliberately does not own `pyproject.toml` — overriding it would make CI test only 3.14 while the package metadata still advertised 3.12 support, trading one incoherence for another.

  Instead the conflict now fails closed. `_reject_conflicting_python_version` runs before the target directory is created and before any prompt, and errors with:

  > `--python-version 3.14 conflicts with the requires-python floor (3.12) declared in pyproject.toml. CI derives its matrix from requires-python, so mise.toml and mypy.ini would pin 3.14 while CI tested 3.12. Set requires-python to >=3.14, or drop --python-version to adopt the declared 3.12.`

  Effective precedence is now: **declared `requires-python` is authoritative; `--python-version` supplies the value only when nothing declares one; a contradicting flag is rejected.** An agreeing flag is accepted. The wizard already skipped its prompt when `requires-python` exists, so the interactive path cannot reach the conflict.

  Covered by `test_flag_contradicting_requires_python_is_rejected` and `test_flag_agreeing_with_requires_python_is_accepted`. The PR body and `docs/guides/using-project-init.md` both claimed the old precedence and are corrected.

Full suite green (2134 passed), `ruff check` clean.

VytCepas:

- **Copilot — `--python-version` accepted with a non-python `--language`**: Fixing. Correct: every `python_floor` consumer is gated on the `python` flag, so `--python-version 3.13 --language go` rendered nowhere and passed silently. Now `_reject_python_version_without_python` errors before the target dir is created: `--python-version 3.13 requires --language python (got --language go); nothing would consume the value.` In the wizard `--language` isn't known at parse time, so the flag is dropped with a visible warning instead of an error. Covered by `test_python_version_without_python_language_is_rejected` and `test_wizard_drops_python_version_when_the_language_is_not_python`.

- **Codex — honor --python-version in CI for existing pyprojects (P2)**: Already addressed in `a60c2ae` (the thread is now outdated). The conflict fails closed rather than half-applying; see the previous review response for why CI was not made to follow the rendered floor.

Full suite green (2136 passed), `ruff check` clean.

VytCepas:

- **Copilot — `--python-version` accepted when `--language` is absent in `--non-interactive`**: Fixing. Correct, and it's the sharper form of the previous finding: `args.language` is `None` when the flag is omitted, so my `language != "python"` guard skipped entirely, and `project-init --non-interactive --python-version 3.13` (no `--language`) exited 0 with `mypy.ini` never rendered. Verified by hand before and after.

  A non-interactive run resolves an absent `--language` to `"none"` in `_resolve_inputs`, so the preflight now validates against that same effective value rather than letting `None` read as "unknown". An interactive run genuinely doesn't know the language at parse time, so it still defers to the wizard's warning path.

  ```
  project-init: error: --python-version 3.13 requires --language python (got --language none); nothing would consume the value.
  ```

  The existing test is now parametrized over `["go", None]` so the absent-language case is pinned.

Full suite green (2137 passed), `ruff check` clean.

PR #716, opened 2026-07-09
fix(PI-715): make the review gate satisfiable by the agents that review

VytCepas:
## Problem

`setup_github.sh --protect` provisioned `required_approving_review_count: 1` on **every** profile. But an approving review is unsatisfiable on a solo repo:

- GitHub refuses self-approval;
- Copilot and Codex submit `COMMENTED` reviews, never `APPROVED`.

So `reviewDecision` never left `REVIEW_REQUIRED`, `review/decision` never went green, and the scaffolder's own `monitor_pr.sh --merge` could only merge through `--admin`. With `enforce_admins: false` the gate was never binding — it just guaranteed that **every** PR ended in an override, and trained the operator to reach for `--no-review`.

The script's own comment already diagnosed this for `review/decision` ("unsatisfiable for a solo owner (you can't approve your own PR)") and then required an approving review anyway.

Confirmed live on this repo: removing the requirement turned `mergeStateStatus` from `BLOCKED` to `CLEAN`, and #713 merged with a plain `--merge` — the first non-admin merge here.

## Changes

**`setup_github.sh`** — `required_approving_review_count` is profile-derived: `0` for `individual`/`standalone`, `1` for `org` (real reviewers exist there). The org-only ruleset keeps its own requirement. `required_conversation_resolution` stays on for every profile — that is what actually enforces "resolve the comments before merging", and unlike an approval, an agent can satisfy it.

**`monitor_pr.sh`** — an empty `reviewDecision` means "no approval policy", and the script used to fall straight through to the merge, i.e. merge with *zero* review. It now:
- waits for at least one review to land (same 6-minute budget as the approval wait);
- refuses to merge while any review thread is unresolved, printing the comments and exiting 2 to open a review cycle;
- never force-merges past open comments — `required_conversation_resolution` would reject the merge anyway, and `--admin` there would silently discard feedback nobody answered.

**`review-status.yml`** — the `review/decision` status now reports what an agent workflow can actually satisfy:

| condition | state |
|---|---|
| `CHANGES_REQUESTED` | `failure` |
| no review yet | `pending` |
| unresolved review threads | `failure` (drives the review cycle) |
| `APPROVED` | `success` |
| reviewed, nothing open | `success` |

Triggers extended to `pull_request_review_thread` (resolve/unresolve) so clearing the last thread flips the check, and to `pull_request: synchronize` so a new commit gets its own status instead of leaving the required check missing on the head SHA.

While here: the root copy carried an env-routing hardening fix the template never received. Both copies are now identical.

**Docs** — the `github_workflow` skill no longer presents `--no-review` as the routine solo path.

## Verification

`tests/integration/test_review_gate.py` drives the real script with stubbed `gh`/`git`:
- unresolved comments → exit 2, review cycle, no merge
- reviewed with no open comments → `Merged PR #1`, and **not** `(admin)`

Plus contract assertions that solo profiles render `REQUIRED_APPROVALS=0` while the org branch and the org ruleset keep `1`.

Byte-identity fixtures: only `review-status.yml` (all four combos) and `github_workflow/SKILL.md` (no_plugin combos, where the skill ships as a file rather than via the plugin) were re-pinned, after verifying no other file drifted. Both test docstrings carry the exception note.

Full suite green (2142 passed), `ruff check` clean, shellcheck gate passes.

Closes #715


VytCepas:

- **Codex — re-read reviewDecision after the no-policy wait (P1)**: Fixing. Confirmed and correct — this was a merge-safety hole I introduced in this very PR. `REVIEW_DECISION` is read once, before any review exists. On a no-approval-policy branch it comes back empty, and the new wait loop only polled `_has_review_activity`. A review landing during that wait as `CHANGES_REQUESTED` — in particular a summary-only change request, which leaves **no** unresolved thread — would leave the variable empty, skip the `CHANGES_REQUESTED` block entirely, pass the `UNRESOLVED=0` gate, and merge the PR over a requested change.

  `REVIEW_DECISION` is now re-queried at the end of the no-approval-policy block, unconditionally rather than only when the script waited — a review may also have landed just *before* the first read returned. The existing `CHANGES_REQUESTED` handler then sees it and opens a review cycle.

  Pinned by `test_changes_requested_after_the_no_policy_wait_blocks_the_merge`, whose `gh` stub returns an empty decision on the first query and `CHANGES_REQUESTED` afterwards, reproducing the real ordering. I verified the test **fails** when the re-read is removed and passes with it, so it is not vacuous.

Full suite green (2143 passed), `ruff check` clean, shellcheck gate passes.

PR #717, opened 2026-07-09
feat(PI-714): make review cycles configurable and explained in the wizard

VytCepas:
## Problem

`monitor_pr.sh` hardcoded `MAX_REVIEW_CYCLES=2` and nothing surfaced it. A scaffolded project inherited a two-cycle review protocol it never agreed to and could not tune.

## Design

The count flows through the existing seams — no `.sh.tmpl` conversion needed, because `config.yaml` already renders variables and `gh_host.sh` already reads `profile` out of it:

```
wizard / --review-cycles → .agents/config.yaml → gh_host.sh review_cycles() → monitor_pr.sh
                                                  PI_REVIEW_CYCLES overrides per-run
```

| Cycles | Meaning |
|---|---|
| `0` | no review control — merge as soon as CI is green |
| `1` | comment once, resolve, merge |
| `2` (default) | the resolved comments get reviewed too, then merge |
| `3+` | additional re-review rounds |

Whatever the count, the merge rule is the same: **once the review agents' comments are resolved — or none arise — the PR is free to merge.**

**`0` is not an admin override.** It skips *this script's* review gate; if an approval policy is in force the merge still blocks. "No review control" is a choice about cycles, not a licence to bypass branch protection. The zero branch is set *before* the max-cycles check, which would otherwise read `0 >= 0` as "cycles exhausted" and force an `--admin` merge — the exact bug this would have shipped.

**Registered as a selectable concern (ADR-023), not a mechanical flag.** `0` trades review control for speed; that cost has to be stated before asking, so it gets an explainer panel, a `WIZARD_CONCERN_FLAGS` entry, and a render check in `test_wizard_explanations.py`.

**Gated on the lifecycle.** A `--lifecycle none` project ships no `monitor_pr.sh`, so there is no gate to size: the wizard never asks, the `review_cycles` key is not rendered into `config.yaml`, and an explicit `--review-cycles` is rejected rather than recorded where nothing can read it.

## Verification

14 tests in `tests/integration/test_review_cycles.py`, plus a live check of the scaffolded output:

- `--review-cycles 0|1|2|5` land in `config.yaml`; default is `2`
- `--lifecycle none` renders **no** `review_cycles:` key and no `monitor_pr.sh`
- `--lifecycle none --review-cycles 2` → rejected, nothing written
- `--review-cycles -1` → rejected
- `gh_host.sh review_cycles()` reads the configured value back, and falls back to `2` for a pre-714 project with no key (not `0`)
- the reader's `sed` cannot match the `review_cycles` value embedded in `config.yaml`'s `variables:` JSON blob
- `PI_REVIEW_CYCLES=0` disables the gate against the real script with a stubbed `gh`
- the explainer states "no review control", `Helps:`, and `Default: 2`
- the wizard does **not** ask when lifecycle is declined

Adding a prompt shifted four scripted-wizard tests; each now stubs `_choose_review_cycles_interactive` with a comment. `tests/helpers.make_variables` gained `review_cycles` — without it the placeholder rendered literally and 342 tests failed.

Full suite green (2158 passed), `ruff check` clean, shellcheck gate passes.

Closes #714

### DIFF
--- a/.agents/scripts/gh_host.sh
+++ b/.agents/scripts/gh_host.sh
@@ -69,6 +69,17 @@ gh_profile() {
   printf '%s\n' "${prof:-individual}"
 }
 
+# Review-fix cycles monitor_pr.sh runs before it stops asking for another pass
+# (#714). 0 means no review control: merge as soon as CI is green. Anchored on
+# this file's location for the same reason gh_profile is. Falls back to 2 — the
+# default the wizard explains — when unset or absent.
+review_cycles() {
+  local cfg n=""
+  cfg="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/../config.yaml"
+  [ -f "$cfg" ] && n=$(sed -nE 's/^[[:space:]]*review_cycles:[[:space:]]*([0-9]+).*/\1/p' "$cfg" | head -1)
+  printf '%s\n' "${n:-2}"
+}
+
 # Base branch for feature PRs. Single trunk: the scaffolder pins the rendered
 # workflows (ci.yml, validate-pr.yml) to 'main', so this MUST return 'main' too —
 # resolving the live default branch instead would let start_issue.sh target a
--- a/.agents/scripts/monitor_pr.sh
+++ b/.agents/scripts/monitor_pr.sh
@@ -48,7 +48,16 @@ PY="$SCRIPT_DIR/../hooks/_py.sh"
 PR_NUMBER="${1:-}"
 MODE=""
 REVIEW_CYCLE=0
-MAX_REVIEW_CYCLES=2
+# #714: review-fix cycles before the script stops asking for another pass.
+# Precedence: PI_REVIEW_CYCLES env > config.yaml `review_cycles` > 2.
+# 0 disables review control entirely — merge as soon as CI is green.
+MAX_REVIEW_CYCLES="${PI_REVIEW_CYCLES:-$(review_cycles)}"
+case "$MAX_REVIEW_CYCLES" in
+'' | *[!0-9]*)
+  echo "review cycles must be a non-negative integer; got '$MAX_REVIEW_CYCLES'" >&2
+  exit 2
+  ;;
+esac
 NO_REVIEW=0
 ALLOW_ADMIN=0
 
@@ -456,6 +465,19 @@ fi
 REVIEW_TIMEOUT=360
 REVIEW_ELAPSED=0
 REVIEW_DECISION=$(_get_review_decision)
+
+# #714: review_cycles=0 means the operator declined review control. Skip the whole
+# review gate — the wait, the changes-requested cycle, the unresolved-thread check
+# — and merge on green CI. Deliberately NOT an admin override: if an approval
+# policy is in force the merge still blocks, because "no review control" is a
+# choice about this script's cycles, not a licence to bypass branch protection.
+# Set before the max-cycles check below, which would otherwise read 0 >= 0 as
+# "cycles exhausted" and force an admin merge.
+if [ "$MAX_REVIEW_CYCLES" -eq 0 ] && [ "$MODE" = "--merge" ]; then
+  echo "PR #$PR_NUMBER: CI passed. review_cycles=0 — no review control; merging on green CI."
+  REVIEW_DECISION="SKIPPED"
+fi
+
 if [ "$MODE" = "--merge" ] && [ "$REVIEW_CYCLE" -ge "$MAX_REVIEW_CYCLES" ] && [ "$REVIEW_DECISION" = "REVIEW_REQUIRED" ]; then
   echo "Max review cycles ($MAX_REVIEW_CYCLES) reached — skipping reviewer wait and force-merging with admin override."
   _admin_merge
--- a/.agents/scripts/monitor_pr.sh
+++ b/.agents/scripts/monitor_pr.sh
@@ -58,6 +58,9 @@ case "$MAX_REVIEW_CYCLES" in
   exit 2
   ;;
 esac
+# Base-10, so a config.yaml or PI_REVIEW_CYCLES carrying "08" can never reach an
+# arithmetic context as octal (PR #717 review).
+MAX_REVIEW_CYCLES=$((10#$MAX_REVIEW_CYCLES))
 NO_REVIEW=0
 ALLOW_ADMIN=0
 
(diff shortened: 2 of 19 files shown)


VytCepas:

Both reviewers found the same real defect, from different angles. Fixing.

- **Codex — validate interactive review-cycle flags before recording them (P2)** and **Copilot — `--review-cycles` unvalidated / silently dropped in the interactive path**: Correct. `_resolve_review_cycles` only ran on the non-interactive path; the wizard copied the raw CLI value. So `project-init <target> --review-cycles -1` (interactive, lifecycle on) wrote `review_cycles: -1` into `config.yaml`, which `monitor_pr.sh` then rejects at parse time — an invalid scaffold produced by a valid-looking command. And `--lifecycle none --review-cycles 2` was dropped without a word.

  Restructured so validation is shared rather than duplicated:
  - `_validate_review_cycles(args, parser)` now runs in `main()` on **both** paths, before the target directory is created and before any prompt. It rejects a negative value and an explicit `--lifecycle none` pairing.
  - `_resolve_review_cycles(args)` is reduced to the non-interactive default resolution, with validation already done.
  - The one case `main()` genuinely cannot know is a tier the user picks *at the prompt*. Aborting a half-answered wizard there would be worse than the bug, so that path prints `--review-cycles N ignored: no merge gate is scaffolded with lifecycle 'none'` and drops the value — mirroring how `--python-version` handles the same situation (#628).

  Four new tests: the wizard warns and zeroes on a chosen `lifecycle none`; the flag still beats the prompt when lifecycle is on; and both rejections fire on an *interactive* invocation with nothing written to the target.

Full suite green (2162 passed), `ruff check` clean.

VytCepas:

- **Copilot — a preset-resolved `lifecycle = "none"` bypasses the gating contract**: Fixing. Correct, and it's the same bug one level up from cycle 1: I validated `args.lifecycle`, but the tier a run actually scaffolds is `_normalize_lifecycle(args.lifecycle) or preset_lifecycle` (`__main__.py:2530`). A company preset that sets `lifecycle = "none"` with no `--lifecycle` flag therefore recorded `review_cycles: 2` for a project that scaffolds no `monitor_pr.sh`, and an explicit `--review-cycles N` sailed through.

  `_resolve_review_cycles` and `_validate_review_cycles` now both take the **effective** tier rather than reading `args`. `main()` computes it and passes it — but only for a non-interactive run: in the wizard the preset merely seeds the prompt's default, so the tier genuinely isn't knowable at parse time and `None` is passed, leaving the warn-and-drop path to handle it. The validation call moved after `_resolve_preset_lifecycle`, still before the target directory exists.

  Three new tests exercise the resolver directly: preset-none zeroes the count; `--lifecycle github` still overrides a preset that says none; an explicit flag against an effective-none tier raises; and an interactive run with `None` does not.

  A note on how I tested it: my first attempt scaffolded with a temp preset directory via a `PROJECT_INIT_PRESETS_DIR` env var. No such override exists, so both tests would have called `pytest.skip` and reported green while asserting nothing. Replaced with direct calls into the resolver.

Full suite green (2165 passed), `ruff check` clean.

VytCepas:

- **Copilot — the `review_cycles=0` branch is gated on `--merge`, contradicting its own comment**: Fixing. Correct. The comment claimed the branch "skips the whole review gate", but `[ "$MODE" = "--merge" ]` meant a monitor-only run (`monitor_pr.sh <pr>`) still fell into the reviewer wait loop and sat there for the full 360s — waiting on a gate the operator had explicitly switched off.

  The condition now keys only on `MAX_REVIEW_CYCLES -eq 0`; the merge/monitor distinction survives only in the message ("merging on green CI" vs "skipping the review gate"). Two guards added: a behavioral test asserting a monitor-only run with `PI_REVIEW_CYCLES=0` prints no "Waiting for reviewer" and exits 0, and a static assertion that the `&& [ "$MODE" = "--merge" ]` pairing does not reappear.

Full suite green (2166 passed), `ruff check` clean, shellcheck gate passes.

VytCepas:

- **Copilot — leading zeros in the cycle count are read as octal**: Fixing, though the stated mechanism is only half right, and the half that's wrong pointed at the wrong variable.

  Measured on bash 5:
  - `[ 1 -ge 08 ]` and `[ "$M" -eq 0 ]` with `M=08` do **not** error — `test`/`[` parse base-10. So `MAX_REVIEW_CYCLES`, which only ever appears in `[ ]` comparisons, was never at risk from the path described.
  - `$((08 + 1))` **does** die with `value too great for base`. The variable that reaches an arithmetic context is `REVIEW_CYCLE`, via `NEXT=$((REVIEW_CYCLE + 1))` at four sites.

  So the real bug is `monitor_pr.sh <pr> --merge --review-cycle 08`: the digits-only `case` guard accepts it, and the script then crashes when it tries to print the next cycle number. Normalized both `REVIEW_CYCLE` and `MAX_REVIEW_CYCLES` with `$((10#…))` immediately after validation — the latter defensively, since `config.yaml` and `PI_REVIEW_CYCLES` are user-editable and a future edit could route it into arithmetic.

  Regression test drives the real script with a stubbed `gh` that returns unresolved comments, which is what carries execution into the `NEXT=$((…))` branch. My first attempt at this test stubbed nothing and would have exited at the `command -v gh` check before parsing ever happened — passing vacuously. I checked: with the `$((10#…))` normalization removed, the test **fails** on both `08` and `09`; with it, both pass.

Full suite green (2168 passed), `ruff check` clean, shellcheck gate passes.

VytCepas:

- **Copilot — `review_cycles:` renders with 3 leading spaces, risking invalid YAML**: **Not applying — measured false.** The `{{#if lifecycle}}` tag opens the line and is consumed by the renderer, so the key lands at exactly the sibling indent. Verified against real scaffolded output:

  ```
  review_cycles indent: 2
  sibling (github_project_number) indent: 2
  ```

  and both the lifecycle-on and lifecycle-off renders round-trip through `yaml.safe_load()`:

  ```
  lifecycle=github -> ok, project.review_cycles = 2
  lifecycle=none   -> ok, project.review_cycles = None   (key absent, as designed)
  ```

  There is one genuinely cosmetic effect: with `lifecycle none` the standalone `{{/if lifecycle}}` line leaves an extra blank line. Valid YAML, consistent with how every other whole-line `{{#if}}` gate in this template behaves, and not worth re-pinning eight byte-identity fixtures to tidy.

All checks green, no outstanding findings. Merging.

PR #718, opened 2026-07-09
chore(PI-636): measure coverage without gating on it

VytCepas:
## Problem

No `pytest-cov`, no `[tool.coverage]`, no Codecov — zero visibility into which code paths and templates the suite actually exercises, despite CLAUDE.md requiring a test for every `templates/` change.

## Changes

- `pytest-cov` and `coverage>=7.10` in the `dev` group (lockfile updated — CI runs `uv sync --locked`)
- `[tool.coverage]` with `branch = true`, `parallel = true`, `patch = ["subprocess"]`, scoped to `src/project_init`
- `just test-cov` for a local report; `--cov --cov-report=term-missing` on the CI test step

## No `fail_under`, deliberately

The issue offered a soft floor as optional. I left it out. A hard floor blocks PRs that touch nothing related to the dip, and this repo's real coverage contract — *"any change to `templates/` should have a corresponding test"* — is not expressible as a percentage. The number is a signal to read, not a gate to satisfy. A contract test pins that decision so it isn't quietly reversed.

## The part that actually mattered: subprocess capture

Most of the integration suite drives the scaffolder as a subprocess (`python -m project_init`), and **coverage records nothing inside a child process** by default.

Measured, before enabling it:

```
src/project_init/__main__.py   823  590  ...  24%
src/project_init/concerns.py   221  221  ...   0%     <- exercised on every CLI run
TOTAL                                        11%
```

`concerns.py` reading as 0% dead code while the CLI tests hammer it is not "low coverage", it is a **wrong** number — worse than no number, because someone would delete live code on the strength of it.

`[tool.coverage.run] patch = ["subprocess"]` (coverage 7.10+) makes coverage instrument the children itself. After enabling it the same suite reports **92% total**, with `scaffold.py` at 96% and `upgrade.py` at 93%.

Verified load-bearing rather than assumed: removing the option drops the suite back to **11%** with `concerns.py` at 0%.

*(An earlier revision of this PR exported `COVERAGE_PROCESS_START` from `tests/conftest.py` instead. That worked — coverage 7.15 ships its own `a1_coverage.pth` which reads the variable — but it depended on another package's `.pth` picking up an env var, which is indirect and fragile. Replaced with coverage's own declarative option per review; no conftest hook remains.)*

## Why `just test` stays coverage-free

`just test` is what agents run for the final green check. Coverage adds ~50s (123s → 173s) and ~15 lines of report to every run — output that lands in agent context each turn (token-efficiency, PI-641). CI reports it instead, where the log is read once by a human.

Bare `--cov` (no `=path`) is deliberate: pytest-cov 7 does not rewrite the configured `source`. Measured — the report covers 10 `src/project_init` files, with zero `tests/` or `tools/` rows. Repeating the path in the justfile *and* the workflow would add two more places to drift from the config, the exact bug class #628 just fixed.

## Verification

Seven contract assertions in `tests/contracts/test_coverage_config.py`: pytest-cov is a dev dep; branch + parallel coverage scoped to the package; `patch = ["subprocess"]` present; the *resolved* coverage version is ≥ 7.10 (asserted via `importlib.metadata`, not by parsing the requirement string); no `fail_under` anywhere; no conftest env hook; CI reports without `--cov-fail-under`.

Full suite green (2150 passed), `ruff check` clean.

Closes #636


VytCepas:

- **Codex — enable coverage's subprocess patch (P2)**: Applying, with one correction to the diagnosis. The premise is right — pytest-cov 7 ships no subprocess `.pth`. But the conclusion that subprocess coverage "still won't be included" was not what I measured: it *was* working, because **coverage 7.15 installs its own `a1_coverage.pth`**, which honors `COVERAGE_PROCESS_START`. Verified: `sysconfig purelib` contains `a1_coverage.pth`, no `pytest-cov.pth`.

  That said, the recommendation is the better design regardless, and I've taken it. Depending on another package's `.pth` picking up an env var I export from `conftest.py` is fragile and indirect. `patch = ["subprocess"]` is coverage's own declarative mechanism, needs no env var, and cannot leak a recorder into `just test`. The conftest hook is **removed**.

  Verified load-bearing rather than assumed — with the option removed, the same suite reports **11%** total and `concerns.py` at **0%**; with it, **92%** and `concerns.py` at 92%. Also pinned `coverage>=7.10` directly, since `patch` is coverage's option, not pytest-cov's, and nothing else guaranteed the floor.

- **Copilot — `getoption()` expects the option dest, not the CLI flag string** (and the matching claim that the contract test "will pass while the guard is actually broken"): **Not applying — the claim is incorrect.** `pytest.Config.getoption()` accepts the option string and strips the leading dashes. Measured directly:

  ```
  pytest_configure: config.getoption("--cov", default=None)
    -> ['src/project_init']        # truthy
  ```

  So the guard fired and subprocess coverage was being collected, which is exactly why the reported total moved from 11% to 92% in the first place. The finding is moot now anyway, since the conftest hook is gone entirely per the Codex finding above — but I'd rather record that it was wrong than let a false claim stand as the reason for a change.

Full suite green (2150 passed), `ruff check` clean.

VytCepas:

Five comments, all from Copilot. Two are correct and fixed; three rest on a premise I measured and found false.

**Fixed**

- **Stale module docstring in `test_coverage_config.py`**: Correct — I rewrote the mechanism from a `COVERAGE_PROCESS_START` conftest hook to `patch = ["subprocess"]` and left the docstring describing the old one, in the very file that asserts the hook is gone. Updated to describe `[tool.coverage.run] patch` and coverage>=7.10, with the concrete failure mode (11% total, `concerns.py` at 0%).

- **Brittle version-floor parsing**: Correct. `pin.split(">=")[1]` breaks on `==7.15.0`, `>=7.10,<8`, extras, or markers. Rather than reach for `packaging` (a transitive dep, not a declared one), the test now asserts the **resolved** version via `importlib.metadata.version("coverage")` — which is what actually decides whether `patch` is honored — and separately that a direct `coverage` entry exists in the dev group. Stronger and unbreakable by spec syntax.

**Not applying — measured false**

- **`pytest --cov` with no argument defaults to `--cov=.` and overrides the configured `source`** (raised three times: `justfile:27`, `ci.yml:54`, and as a proposed contract assertion): This is not the behavior of pytest-cov 7.1.0. Bare `--cov` leaves `[tool.coverage.run] source` in force. Measured on this branch:

  ```
  $ uv run pytest tests/unit -q --cov --cov-report=term
  # rows matching ^src/project_init/ : 10
  # rows matching ^(tests|tools)/    : 0
  # TOTAL stmts 2575  == the package alone
  ```

  If the flag were rewriting the source to `.`, `tests/` and `tools/` would appear in the table and the statement count would be several times larger. They don't, and it isn't.

  I've kept bare `--cov` deliberately: the scope belongs in `[tool.coverage.run]`, and repeating `src/project_init` in the justfile *and* the workflow would create two more places to drift from the config — the exact class of bug this repo keeps hitting (see the three-way Python-pin drift in #628). The contract test that pins `source = ["src/project_init"]` is what makes bare `--cov` safe, and I've documented that reasoning in `test_ci_reports_coverage_without_gating_on_it` so the next reader doesn't re-litigate it.

Full suite green (2150 passed), `ruff check` clean.

VytCepas:

- **Copilot — the PR description contradicts the implementation**: Correct, and a fair catch. The description still credited a `tests/conftest.py` export of `COVERAGE_PROCESS_START`, while the code (and `test_no_conftest_env_hook_is_needed`) asserts no such hook exists — I rewrote the mechanism to `[tool.coverage.run] patch = ["subprocess"]` after Codex's review and updated the code, tests, and docstring but not the PR body.

  The description now describes the implemented approach, with the abandoned one recorded as a parenthetical so the review history stays legible rather than silently erased. No code change: the tests already pin the real mechanism, including that the conftest hook is gone.

All checks green, no outstanding findings. Merging.

VytCepas:

- **Copilot — contract test reads `pyproject.toml` / `tests/conftest.py` / `ci.yml` via cwd-relative paths (×3)**: Fixing. Correct. Several contract tests in this repo already anchor on `__file__` (`test_lifecycle_byte_identity.py`, `test_agent_overlays.py`, `test_marketplace_source.py`), so this was inconsistent as well as fragile. All three now resolve from a `_ROOT = Path(__file__).resolve().parents[2]`.

  Verified from a subdirectory rather than assumed: `cd src && pytest ../tests/contracts/test_coverage_config.py` passes 7/7, where the relative-path version would have raised `FileNotFoundError`.

PR #720, opened 2026-07-10
fix(PI-719): review-status.yml never loaded — invalid Actions trigger

VytCepas:
## Problem

PR #716 (PI-715) added `pull_request_review_thread` as a trigger to `review-status.yml`. **It is not a valid GitHub Actions trigger** — it exists as a webhook event but is absent from the events-that-trigger-workflows list. GitHub rejects the entire workflow file at load time.

Every run since #716 merged has been a **startup failure**: zero jobs, zero check-runs, no `review/decision` status, and the workflow's display name degraded to its file path.

```
b88fe125 push failure 2026-07-10T06:54:30Z
4596e82c push failure 2026-07-09T23:05:52Z   <- first failure, right after #716
--------------- before #716 ---------------
a55b047b pull_request_review success 2026-07-09T17:33:32Z
```

Confirmed three ways: `gh run view` says "This run likely failed because of a workflow file issue"; the runs have no jobs and no check-runs; `actionlint` reports `unknown Webhook event "pull_request_review_thread"`. GitHub's current docs list `pull_request`, `pull_request_review`, `pull_request_review_comment`, and `pull_request_target` — no `_thread`.

It went unnoticed for a day because `review/decision` is not a required check on `main`, and `monitor_pr.sh` queries `reviewDecision` and review threads through the API rather than reading the commit status. **I introduced this in #716 and did not catch it**, because the tests only asserted on the file's text and `yaml.safe_load()` parses it fine.

## Fix

- **Drop the invalid trigger.** Resolving a review thread emits *no* workflow event, so #716's "clearing the last thread flips the check green" is unachievable as designed. `pull_request_review_comment` and `issue_comment` (guarded with `github.event.issue.pull_request`) refresh the status on the comment activity that accompanies a resolution round.
- **Resolve the head SHA from the API**, not the payload. Only `pull_request*` events carry `.pull_request.head.sha`; on `issue_comment` the status would have been posted to an empty SHA.
- **Add an actionlint gate** — in CI and as `tests/integration/test_workflow_lint_gate.py` — over this repo's workflows *and* the workflows a scaffold renders, so a broken trigger can neither land here nor ship to a new project.

## Verification

- `actionlint` clean over all root workflows and all rendered scaffold workflows.
- The gate is **non-vacuous**: reintroducing `pull_request_review_thread` makes both tests fail; removing it makes them pass.
- Byte-identity fixtures: only `.github/workflows/review-status.yml` re-pinned, after verifying no other file drifted.
- Full suite green (2177 passed), `ruff check` clean.

## Consequence for #715's goal

`review/decision` cannot be made a required status check until this lands. Even after it lands, a thread resolved with no accompanying comment, review, or push will not refresh the status — worth knowing before anyone makes it required.

Closes #719

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,17 +157,6 @@ jobs:
           printf '%s\n' "${files[@]}"
           uvx --from 'shellcheck-py==0.11.0.1' shellcheck --severity=warning "${files[@]}"
 
-      # PI-719: an invalid trigger name made GitHub reject review-status.yml at
-      # load time — every run a startup failure with no jobs, for a full day,
-      # because YAML parsed fine and nothing linted the workflow schema. Lint
-      # this repo's workflows AND the rendered scaffold's, so a broken trigger
-      # can neither land here nor ship to a new project. Pinned for determinism.
-      - name: actionlint workflow files (this repo + rendered scaffold)
-        run: |
-          set -euo pipefail
-          uvx --from 'actionlint-py==1.7.12.24' actionlint \
-            .github/workflows/*.yml /tmp/sc-target/.github/workflows/*.yml
-
   # PI-363 Layer 3 gate: only spend the 10x-billed macOS minutes when shell or
   # template files actually change. Ordinary PRs skip the macOS job entirely.
   changes:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,6 +157,17 @@ jobs:
           printf '%s\n' "${files[@]}"
           uvx --from 'shellcheck-py==0.11.0.1' shellcheck --severity=warning "${files[@]}"
 
+      # PI-719: an invalid trigger name made GitHub reject review-status.yml at
+      # load time — every run a startup failure with no jobs, for a full day,
+      # because YAML parsed fine and nothing linted the workflow schema. Lint
+      # this repo's workflows AND the rendered scaffold's, so a broken trigger
+      # can neither land here nor ship to a new project. Pinned for determinism.
+      - name: actionlint workflow files (this repo + rendered scaffold)
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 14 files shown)


VytCepas:

- **Copilot — the gate skips in CI, and the skip reason claims otherwise**: Correct on both counts, and the second half is the worse half. `lint-and-test` never installed `actionlint`, so the module would have skipped in CI while its own skip message read *"CI installs it"* — a gate that reports green while testing nothing, which is exactly the failure mode this PR exists to fix.

  `actionlint-py` is now a declared **dev dependency**, so `uv sync --group dev --locked` puts the binary on PATH and the gate always runs — in CI and locally. Confirmed: a plain `uv run pytest` collects and runs `tests/integration/test_workflow_lint_gate.py ..` rather than skipping it. The skip guard now only fires for someone running pytest outside the project environment, and says so.

  I also **removed** the separate actionlint step I had added to the `shellcheck` CI job. It was redundant once the test covers both this repo's workflows and the rendered scaffold's — and it was failing, which is its own finding:

- **Self-reported (CI, not a reviewer):** that step failed on this very PR. `actionlint` shells out to `shellcheck` and `pyflakes` when they are on `$PATH`. GitHub runners preinstall shellcheck; my machine does not. So the step passed locally and failed in CI on `run:`-block *style* findings (`SC2016:info`, `SC2086:info`, `SC2034:warning`) in `board-automation.yml` and `ci.yml` — files this PR never touched.

  The gate exists for workflow **schema** validity (triggers, expressions, contexts), which is what actually broke in #716. `-shellcheck= -pyflakes=` disable the embedded linters; the scaffolded shell scripts still get a real shellcheck run in the job above. Reproduced the CI condition locally by putting `shellcheck` on PATH, confirmed the failure, then confirmed the flags fix it — and re-confirmed the gate still fails when the invalid trigger is reintroduced.

Full suite green (2177 passed), `ruff check` clean, all workflows actionlint-clean.

PR #724, opened 2026-07-10
feat(PI-686): enforce staleness for the scaffolded CODE_MAP and own skills INDEX

VytCepas:
## Re-verified before implementing — half the issue was stale

Posted in full as an [issue comment](https://github.com/VytCepas/project-init/issues/686#issuecomment-4933274562):

| Item | Status |
|---|---|
| 1. `.agents/CAPABILITIES.md` staleness | **Moot.** The file does not exist, and `.gitignore:36-46` ignores `.agents/*` except an allowlist it is not on. It could not be committed even if regenerated. |
| 4. "this repo has no committed, current CODE_MAP" | **False.** It exists, is tracked, and is *already* enforced by `test_agents_template_sync.py:51`. |
| 2. Own skills INDEX coverage | **Real.** Implemented here. |
| 3. Scaffolded CODE_MAP staleness | **Real, and the substantive one.** Implemented here. |

## 3. Scaffolded CODE_MAP staleness

`gen_code_map.py` was invoked only by the `code-map` recipe in `justfile.tmpl:103`. Verified absent from the scaffolded `ci.yml.tmpl`, `pre-push.tmpl`, `install_hooks.sh`, and `dot_pre-commit-config.yaml.tmpl` — **no** regen-and-diff step anywhere. So the file whose own header says *"read this before grepping"* rotted freely in every scaffolded project, routing agents to a stale index. That is worse than having no map: an agent trusts it.

CI now regenerates into a temp dir and compares bytes. The generator writes `.agents/docs/CODE_MAP.md` **relative to cwd**, so running it from a temp dir produces a clean copy without touching the committed file — the same trick `test_committed_code_map_is_fresh` uses. It **skips cleanly** when the map doesn't exist, so a fresh scaffold is not born red before its first `just code-map`. Python-gated, since the generator is a Python AST walker.

**Chose CI over the pre-commit hook**, which the issue suggested as the cheapest home. The hook would have to compare the *staged* snapshot: the existing lint gate does an unstaged-patch dance to isolate the index, and a code-map check running after that restore would compare the working tree, so unrelated unstaged WIP would report a false "stale map". CI checks out exactly the commit, so the comparison is unambiguous.

## 2. Own skills INDEX coverage

`test_skill_index.py`'s `_INDEX_PATH` points at `templates/fallback/dot_agents/skills/INDEX.md.tmpl` — the **template**. The only assertion on this repo's own `.agents/skills/INDEX.md` was `test_wiki_skill.py:143-152`, checking for the literal string `"wiki"`. A newly added own-repo skill missing from INDEX passed CI. (All 7 current dirs *are* listed — the gap was the absent guard, not present drift.)

## Verification — behavior, not just presence

Ran the CI step's exact shell against a real scaffolded project:

| Scenario | Result |
|---|---|
| map current | `is current.` exit 0 |
| public function added, map not regenerated | **exit 1**, stale detected |
| after `just code-map` | exit 0 again |
| fresh scaffold, map never generated | **skipped**, exit 0 — not born red |

The INDEX guard was verified non-vacuous by dropping an unlisted skill dir into `.agents/skills/` and watching it fail.

Contract tests assert the step ships for python and is **absent** for a go scaffold. Rendered scaffold workflows stay actionlint-clean (the #719 gate).

Full suite green (2184 passed), `ruff check` clean.

Closes #686

### DIFF
--- a/templates/base/dot_github/workflows/ci.yml.tmpl
+++ b/templates/base/dot_github/workflows/ci.yml.tmpl
@@ -158,6 +158,35 @@ jobs:
       - name: Typecheck
         run: just typecheck
 
+      # PI-686: CODE_MAP.md's header tells agents to "read this before grepping",
+      # but nothing checked it was current — the generator shipped wired only to
+      # a manual `just code-map`, so the map rotted silently and agents were
+      # routed to a stale index. Regenerate into a temp dir (gen_code_map.py
+      # writes .agents/docs/CODE_MAP.md relative to cwd) and compare bytes;
+      # never rewrite the committed file in CI. Skips cleanly before the first
+      # `just code-map` run, so a fresh scaffold is not born red.
+      - name: CODE_MAP is current
+        run: |
+          set -euo pipefail
+          map=".agents/docs/CODE_MAP.md"
+          gen=".agents/scripts/gen_code_map.py"
+          if [ ! -f "$map" ] || [ ! -f "$gen" ]; then
+            echo "no $map yet — generate it with 'just code-map' and commit."
+            exit 0
+          fi
+          scan="src"
+          [ -d "$scan" ] || scan="."
+          tmp="$(mktemp -d)"
+          trap 'rm -rf "$tmp"' EXIT
+          ( cd "$tmp" && uv run --no-project python \
+              "$GITHUB_WORKSPACE/$gen" "$GITHUB_WORKSPACE/$scan" >/dev/null )
+          if ! cmp -s "$tmp/$map" "$map"; then
+            echo "::error file=$map::CODE_MAP.md is stale — run 'just code-map' and commit"
+            diff -u "$map" "$tmp/$map" | head -n 40 || true
+            exit 1
+          fi
+          echo "$map is current."
+
       # Coverage gate (PI-138/PI-569): always on, not conditional — pytest-cov
       # is pulled in ephemerally via `uv run --with`, the same pattern mypy
       # (PI-558) and mutmut (PI-563) use, so there's no dev-dependency to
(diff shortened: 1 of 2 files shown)


VytCepas:

- **Copilot — the skip guard conflates "missing CODE_MAP" with "missing generator"**: Fixing. Correct, and the consequence is the one this PR exists to prevent: a broken scaffold would print "no CODE_MAP yet" and exit 0, so the staleness check would silently never run — a green check testing nothing.

  The two conditions mean different things. `gen_code_map.py` ships with **every** python scaffold (it is inside the `{{#if python}}` gate, same as this step), so its absence is a broken scaffold and now **fails** with a `::error file=` annotation. A missing map means "not generated yet", which is legitimate on a fresh scaffold and still **skips** — a new project must not be born red before its first `just code-map`.

  Verified behaviorally against a real scaffolded project, not just by reading the diff:

  | State | Result |
  |---|---|
  | generator deleted | **exit 1** — "the CODE_MAP staleness check cannot run" |
  | map absent, generator present | exit 0 — skip |
  | map current | exit 0 |
  | public function added, map not regenerated | exit 1 — stale |

  `test_missing_generator_fails_rather_than_skipping` pins the split, including an assertion that the old `[ ! -f "$map" ] || [ ! -f "$gen" ]` pairing does not come back.

Full suite green (2185 passed), `ruff check` clean, rendered workflows actionlint-clean.

VytCepas:

- **Copilot — the own-repo INDEX guard matches the bare directory name, so prose can create a false pass**: Fixing. Correct. `wiki` is precisely the name that would slip through — a sentence mentioning wikis, or any longer word containing a skill name, would satisfy a substring check while the skill went unindexed. The guard would then report green on the exact drift it exists to catch.

  It now requires the link form `.agents/skills/<dir>/SKILL.md`. Checked first that this is what the INDEX actually uses — all 7 skill dirs are already linked that way, so tightening is strictly stronger and changes nothing today.

  Verified non-vacuous, both directions, against the failure mode named in the review:

  | INDEX contains | Result |
  |---|---|
  | `The zz_probe skill is great.` (prose only) | **fails** — correctly reports it unindexed |
  | `- [zz_probe](.agents/skills/zz_probe/SKILL.md)` | passes |

  (For the record, there are no substring collisions among the current dir names — the risk was prose, not collision.)

PR #728, opened 2026-07-10
fix(PI-726): add a real format gate, and stop rules docs promising one

VytCepas:
## Problem

Every `format` recipe **writes**; none **checked**. An unformatted file merged green for python, node and rust.

And `templates/base/dot_agents/rules/rust.md:18` documented a gate that did not exist:

```
cargo fmt --check    # verifies only; `cargo fmt` (no flag) writes changes
```

Nothing ran it — not the justfile, not CI. `go.md:18` presented `golangci-lint fmt` (which *writes*) as the format command, without mentioning the thing that actually checks. **This is #688's map-not-territory failure caught in the wild**: a hand-written doc restating a gate no code enforced.

## Go was never broken — verified, not assumed

`.golangci.yml:39-42` enables `gofumpt` under `formatters:`. The docs do **not** state whether `golangci-lint run` reports formatter findings or whether they only apply via `golangci-lint fmt` — I read both the config reference and the formatters page, and neither says. So I downloaded golangci-lint 2.1.6 and a Go toolchain and ran it:

```
A) unformatted-but-valid Go   -> 1 issues: * gofumpt: 1    EXIT=1
B) after `golangci-lint fmt`  -> 0 issues.                 EXIT=0
```

Go's format gate is real. It gets **corrected docs, not a redundant fourth check**.

## What `just lint` now runs

| Language | Check | Verified |
|---|---|---|
| python | `ruff format --check .` | exit 1 unformatted / 0 clean |
| node | `bunx @biomejs/biome format .` | exit 1 unformatted / 0 clean |
| rust | `cargo fmt --check` | **not run locally** — no rust toolchain; it is the exact command `rust.md` already documented |
| go | (none added) | gofumpt in `golangci-lint run`, proven above |

Node uses `biome format` without `--write`, **not** `biome ci`: the latter also runs Biome's linter, duplicating eslint, which owns the lint gates here.

## The prerequisite the gate exposed

Adding the check immediately showed that **three shipped `.py` files were not ruff-format clean**: `dag_workflow.py`, `governance_gate.py`, `usage_report.py`. So a fresh python scaffold would have **failed its own lint gate on day one** — precisely the #698 bug, shipped again.

Formatted them. A contract test now fails if any shipped template drifts out of format, so this cannot recur silently. Verified non-vacuous: appending `x   =    1` to a shipped template makes it fail.

Confirmed a fresh scaffold passes: `6 files already formatted, EXIT=0`.

## Docs corrected

`rust.md`, `go.md`, `python.md`, `typescript.md` now describe the gate that exists — which recipe writes, which one checks, and that CI runs it.

## Verification

- 6 new contract assertions in `TestFormatGate`, incl. that `format` recipes still *write* and that Go gains no redundant step.
- Byte-identity fixtures: only `.agents/hooks/dag_workflow.py` re-pinned across all four combos, after verifying no other file drifted. Both test docstrings carry the exception note.
- Full suite green (2191 passed), `ruff check` clean, shellcheck gate passes.

## Follow-up, not fixed here

**This repo itself is not `ruff format` clean** — running it rewrites ~80 unrelated files, so its own `just lint` still has no format check while the scaffold it ships does. Same irony as #637. Worth its own issue.

Closes #726

### DIFF
--- a/.agents/hooks/dag_workflow.py
+++ b/.agents/hooks/dag_workflow.py
@@ -31,6 +31,7 @@ Issue-ref prefix detection:
 
 stdlib only.
 """
+
 from __future__ import annotations
 
 import argparse
@@ -168,7 +169,13 @@ def check_ci_green() -> tuple[bool, str]:
         # miscounted as "still running" forever — and a red one as pending —
         # while the CheckRun `conclusion`/`status` fields stay authoritative.
         state = (entry.get("state") or "").upper()
-        if conclusion in {"FAILURE", "TIMED_OUT", "CANCELLED", "ERROR", "ACTION_REQUIRED"} or state in {
+        if conclusion in {
+            "FAILURE",
+            "TIMED_OUT",
+            "CANCELLED",
+            "ERROR",
+            "ACTION_REQUIRED",
+        } or state in {
             "FAILURE",
             "ERROR",
         }:
@@ -527,7 +534,9 @@ def cmd_push(branch: str | None, max_retries: int, *, force: bool = False) -> in
         # running push_branch.sh while on main bypasses the direct-push guard,
         # since the internal `git push` subprocess is invisible to the
         # PreToolUse hook (PI-202).
-        sys.stderr.write("push: refusing to push main/master directly — open a feature branch + PR\n")
+        sys.stderr.write(
+            "push: refusing to push main/master directly — open a feature branch + PR\n"
+        )
         return 1
 
     code, sha_out = _git(["rev-parse", branch])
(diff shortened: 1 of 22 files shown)


VytCepas:

Four comments. **Three rest on a claim about Biome that is false**, and acting on them would have broken the gate. One is correct and is fixed.

## Not applying (×3) — `biome format .` is already a non-writing check

Copilot asserts, in three places, that `bunx @biomejs/biome format .` "is likely writing changes" and that the fix is `format --check`.

Measured against biome 2.x, on an unformatted `bad.ts`:

```
$ sha256sum bad.ts   # before
$ bunx @biomejs/biome format .
  exit=1   file changed: NO
```

It exits **1** and leaves the file **byte-identical**. Without `--write`, `format` *is* the check — that is exactly why I chose it.

And the suggested flag does not exist:

```
$ bunx @biomejs/biome format --check .
Error: `--check` is not expected in this context
```

So applying this review comment would have turned a working gate into a hard error on every `just lint` for every scaffolded TS project. This is the second time in this PR that the documented behavior of a formatter could not be taken on faith — I verified Go's `golangci-lint run` empirically for the same reason.

I have recorded the finding **at the call site** so nobody "fixes" it later: the justfile comment and `typescript.md` now both say the no-flag form is the check and that biome rejects `--check`.

I did tighten the contract test in the spirit of the third comment: it now asserts `--write` is **absent** from the lint recipe, which is the property that actually matters (the gate must not mutate). Asserting the literal string `format --check` would have pinned a command that cannot run.

## Applying (×1) — the `rc == 2` skip could hide a broken gate

Correct, and the sharpest comment of the four. `ruff` is a declared dev dependency, so a non-zero exit from `python -m ruff` is either real drift or a broken invocation — never a reason to skip. A skip would let the shipped-templates format guard **silently disappear**, which is precisely the failure mode this PR exists to prevent. Removed; it now fails loudly with the exit code and output.

Full suite green (2191 passed), `ruff check` clean.

PR #730, opened 2026-07-10
docs: record two earned conventions, and ship them to scaffolded projects

VytCepas:
## Context

Two conventions this session paid for, plus a correction about where they belong.

## The conventions

**1. Verify the premise before you plan against it.** An issue body, an audit note, or a tool's documented behavior is a *claim*. Checking it first was the cheapest step in every task this week:

| Claim | Reality |
|---|---|
| `@anthropic-ai/claude-code` is pinned but missing from the manifest (#689) | never pinned at all |
| `.agents/CAPABILITIES.md` is committed debris (#686) | does not exist, and is gitignored |
| Go has no format check (#726) | `golangci-lint run` exits 1 on unformatted code |
| `biome format` writes; use `--check` (PR #728 review) | it doesn't write, and `--check` is not a flag biome accepts |
| `typescript-eslint` works with whatever `bun add -d typescript` installs (#732) | TS 7 makes eslint exit 2 |

**2. A test that cannot fail is worse than no test.** Both shipped before being caught:
- PR #689: a line filter skipped any line containing `"pinned"` — which is every line of `PINNED_CCR_VERSION`. It passed with the literal reintroduced.
- PR #724: the CODE_MAP guard OR-ed "missing map" with "missing generator", so a broken scaffold exited 0.
- #732: every TS test asserted `eslint.config.mjs`'s *text*; none ran eslint, so a crashing toolchain passed them all.

## The correction (raised in review)

I first wrote both rules into **this repo's CLAUDE.md only**. That helps exactly one project.

`templates/` is the product. The rules now live in `templates/base/AGENTS.md.tmpl` — the canonical file for scaffolded projects (their `CLAUDE.md` is a redirect) — so every generated project gets them.

CLAUDE.md also gains a **"templates/ is the product"** section, naming the corollary that keeps biting: **this repo is not always a good example of its own output.** It has no `ruff format` gate while the scaffold it ships does (#726), and no `pip-audit` step while every scaffolded project runs one. Don't infer template behavior from this repo's config — read `templates/`, or scaffold into a temp dir.

## On "Karpathy's instructions"

Prompted by the question of whether to adopt them. Worth recording what the research found: **Karpathy did not write that file.** His X thread (2026-01-26) is a *diagnosis* — models make silent assumptions, over-complicate, edit unrelated code. The viral four-principle `CLAUDE.md` is community-authored (Jiayuan Chang, the following day). The only real eval (Augment Code, 40 PRs, self-disclosed as non-independent) shows **flat code quality** and 3–10% efficiency gains; the widely-quoted "41% → 11% errors" figure is an unsourced marketing headline.

Three of its four principles are already enforced here by mechanism rather than instruction — the lifecycle DAG, the byte-identity fixtures, the test gates. The part with residual value is "think before coding", which is what convention 1 states, in this repo's own terms, with its own scars as evidence.

## Verification

Contract test asserts both conventions reach a scaffolded `AGENTS.md` and that `CLAUDE.md` still redirects. Verified non-vacuous: renaming the rule in the template fails the test.


VytCepas:

Both correct. Fixed.

- **Copilot — saying `.agents/CAPABILITIES.md` "does not exist" is misleading**: Right, and the irony is that the sentence sat inside the rule about not asserting claims loosely. `src/project_init/capabilities.py` generates it into **every scaffold** — I confirmed it is present in a minimal `--language go --memory none --lifecycle none` output. What is true, and what #686's correction was about, is that it is not tracked in *this repo's* checkout and could not be even if regenerated, because `.gitignore:36-46` allowlists `.agents/{skills,hooks,scripts,settings.json,agents,config.yaml,docs/CODE_MAP.md}` and nothing else. Reworded to say exactly that.

- **Copilot — `test_claude_md_still_redirects_to_agents_md` only asserts the substring `AGENTS.md`**: Right, and it is the weak-guard shape this very PR names. A passing mention would satisfy it while CLAUDE.md stopped being an entrypoint — leaving the conventions in a file nothing points at. It now asserts the redirect sentence itself (`Canonical agent instructions live in [AGENTS.md](AGENTS.md)`) plus `source of truth`.

  Verified non-vacuous: replacing the sentence with "See also AGENTS.md" fails the test; restoring it passes.

PR #731, opened 2026-07-10
feat(PI-729): give TypeScript a blocking security lint, and unbreak its toolchain

VytCepas:
## Problem

TypeScript shipped **zero blocking security gates**. Python runs ruff's `S` (bandit-derived) rules on every `just lint`. TS's only SAST was the `semgrep` CI job, which is **not** in `ci-gate.needs` — advisory. So a fresh TS repo could merge with OWASP-class findings and nothing caught them.

## The trap this PR is really about

Installing `eslint-plugin-security` naively **does not create a gate**. Its recommended preset sets rules to `warn`, and **eslint exits 0 on warnings**:

```
$ eslint .        # eval(userInput), preset defaults
  3:10  warning  eval with argument of type Identifier  security/detect-eval-with-expression
✖ 1 problem (0 errors, 1 warning)
EXIT=0            <-- merges green
```

Pinned to `error`:

```
  3:10  error  eval with argument of type Identifier          security/detect-eval-with-expression
  8:3   error  Unsafe assignment to innerHTML                 no-unsanitized/property
✖ 2 problems (2 errors, 0 warnings)
EXIT=1            <-- blocked
```

`--max-warnings=0` was not an option: `tsdoc/syntax` is deliberately a warning here. So the six security rules are pinned explicitly.

I also **pinned the type-aware security rules** — `no-floating-promises`, `no-misused-promises`, the `no-unsafe-*` family. They were inherited from `strictTypeChecked`, so an upstream preset change could have silently downgraded them without a single test noticing.

## Blocker discovered while verifying: the TS toolchain was already broken (#732)

`just setup` installed **unpinned** `typescript`. Today that resolves to **7.0.2** (the native rewrite), which `typescript-eslint` cannot parse:

```
$ bun add -d typescript       # -> 7.0.2
$ eslint .
TypeError: Cannot read properties of undefined (reading 'Cjs')
    at @typescript-eslint/typescript-estree/.../shared.js:59
EXIT=2                        <-- crash, not a lint failure

$ bun add -d typescript@5     # -> 5.9.3
$ eslint .
EXIT=0
```

**Every fresh node scaffold's `just lint` was broken** — and with it the CI Lint step and the fail-closed pre-commit hook. Pinned to `typescript@^5`.

Nothing caught this because **every existing test asserts the *text* of `eslint.config.mjs`**. None ever ran eslint. A config can name `"security/detect-eval-with-expression": "error"` while the toolchain it configures crashes on startup and lints nothing.

## So this ships a behavioral test

`tests/integration/test_ts_security_gate.py` scaffolds a node project, `bun add`s exactly the dev-dep set `just setup` installs, runs the **scaffolded** eslint config, and asserts exit codes:

| Input | Expected | Actual |
|---|---|---|
| `eval(userInput)` + `el.innerHTML = dirty` | exit 1 | ✅ 1, both at `error` |
| clean code | exit 0 | ✅ 0 |
| no sources at all (fresh scaffold) | exit 0 | ✅ 0 — not born red |
| `typescript@5` installed, not 7 | — | ✅ |

**Honest limitation, stated in the module docstring:** this repo's CI does not install bun, so these tests **skip there**. A skipped test is not a gate. They run locally (4 passed) and document exactly how the gate was verified. Closing that gap properly means installing bun in CI — worth its own issue, not silently pretending here.

## Also

- `rules/typescript.md` now says which rules are pinned vs inherited, and why the pins exist.
- 15 contract assertions pin every rule at `error` and the `typescript@^5` constraint.

Full suite green (2210 passed), `ruff check` clean.

Closes #729
Closes #732


VytCepas:

Five comments. **All five are correct**, and the Codex P1 is the most important finding on this PR — it breaks every *upgraded* Node project, not just a hypothetical one.

## Codex P1 — new plugins are never installed on an upgraded project

Reproduced exactly as described. A project with an existing `package.json`/lockfile takes the `bun install` branch, which cannot add a dependency the manifest never listed. `eslint.config.mjs` then fails to import:

```
# package.json from the OLD toolchain, new eslint.config.mjs
$ bunx eslint .
Oops! Something went wrong! :(
EXIT=2            <-- crash, before any gate runs
```

Two fixes, because the two audiences differ:

1. **CI** now seeds the toolchain when it is incomplete. `just setup` is idempotent, so the guard is cheap:
   ```sh
   if ! bun pm ls 2>/dev/null | grep -q "eslint-plugin-security"; then
     echo "lint toolchain incomplete (upgraded project?) — running just setup"
     just setup
   fi
   ```
2. **Local runs** get a migration note (`MIGRATION_NOTES["1.0.2"]`, `action_required`): *"Node projects: run `just setup` after upgrading."* `project-init upgrade` surfaces it. Note this activates when the version is bumped to 1.0.2 — notes are keyed by release, per the module's convention.

Verified end to end: **exit 2 → guard fires → exit 1** on `eval(userInput)`.

## Copilot — plugins not registered in the TS block

Correct, and it defeats the purpose of pinning: a rule whose plugin was registered by `security.configs.recommended` still depends on that preset's shape. Now `plugins: { tsdoc, security, "no-unsanitized": noUnsanitized }`.

I checked that flat config **accepts** the re-registration rather than erroring on a redefined plugin — it does, and the gate still exits 1.

## Copilot — `node_modules/.bin/eslint` is not portable

Correct, and inconsistent with what `just lint` runs. Now `bun x eslint .`, which is the same entrypoint and works on Windows.

## Copilot — stderr missing from the failure message

Correct, and specifically bad here: an eslint **crash** (exit 2 — the #732 regression) prints its stack to stderr, so the one failure mode this module exists to detect was the one it reported least legibly. The message now names exit 2 explicitly and includes stderr.

## Codex P2 — skipping on `bun add` failure hides regressions

Correct. A dependency conflict, a bad package name, or an unsatisfiable range is exactly the regression these tests exist to catch; skipping there gives a green run. The fixture now skips **only** for a network failure (`getaddrinfo`/`ENOTFOUND`/`ECONNREFUSED`/`network`/`timed out`) and raises for anything else.

Full suite green (2212 passed), `ruff check` clean, rendered workflows actionlint-clean, and the 4 behavioral gate tests still pass locally.

VytCepas:

- **Copilot — the trailing `\` in the `setup` recipe "is likely to produce a shell parse error"**: **Not applying — measured false.** `just` joins backslash-continued recipe lines into a single command before handing it to the shell. Verified against the **rendered scaffolded justfile**, not a toy:

  ```
  $ just -n setup
  # typescript@^5: typescript-eslint cannot parse with TS 7 (the native rewrite);
  # unpinned, `bun add -d typescript` resolves to 7.x and eslint crashes with exit 2 (#732).
  bun add -d eslint "typescript@^5" typescript-eslint eslint-plugin-jsdoc eslint-plugin-tsdoc eslint-plugin-security eslint-plugin-no-unsanitized @biomejs/biome
  EXIT=0
  ```

  One command, correctly joined. Rewriting it as a single 160-character line, or wrapping it in `sh -c`, would be a change for the worse on a false premise.

  Since a reviewer has now tried to "fix" this once, I pinned the property rather than leave it to prose: `test_setup_recipe_is_a_single_valid_command` asserts the recipe's first executable line starts with `bun add -d` and that exactly **one** `bun add` invocation exists, however the line is wrapped. Verified non-vacuous — splitting it into two `bun add` calls fails the test. (My first version of that assertion counted `bun add` across the whole segment, including the recipe's own comment, and failed. Fixed to count executable lines.)

- **Copilot — comment says `bunx eslint .` while the code runs `bun x eslint .`**: Correct, and worth being precise about since they are the same thing. Reworded: `bun x eslint .` is the subcommand form of `bunx`, and it is exactly what the scaffolded `just lint` runs.

VytCepas:

Both correct. Fixed.

- **Copilot — `_DEV_DEPS` omits `@biomejs/biome` despite claiming to be "exactly what `just setup` installs"**: Right, and the comment made the omission worse than the omission itself: a drifting copy of the toolchain means the behavioral tests exercise something the scaffold does not ship. Added.

  Rather than fix the list and leave the claim on trust, the claim is now a test: `test_dev_deps_match_the_scaffolded_setup_recipe` scaffolds a node project, parses the `setup` recipe out of the rendered justfile (skipping its comments, honoring the `\` continuation), and asserts set equality with `_DEV_DEPS`. Verified non-vacuous — removing `@biomejs/biome` from `_DEV_DEPS` fails it.

- **Copilot — `assert "  error  " in result.stdout` is brittle and redundant**: Right on both counts. `returncode == 1` already proves severity, because eslint exits **0** when every finding is a warning — which is the exact defect this PR fixes. Asserting the human formatter's column spacing on top of that pins eslint's output format for no gain. Removed; the reasoning is now a comment where the assertion was.

Full suite green, `ruff check` clean, all 5 behavioral gate tests pass locally.

VytCepas:

- **Copilot — the helper claims to run "exactly what `just lint` runs" but invokes `bun x` while the recipe uses `bunx`**: Fixing, and taking the stronger of the two options offered. Last cycle I fixed the *comment*; that left the test one alias away from the thing it claims to verify, and this module exists precisely because "asserted it, didn't run it" is how #732 shipped.

  The helper now invokes `bunx eslint .` — byte-for-byte `templates/base/justfile.tmpl:126`. The skip probe also checks for `bunx`, not just `bun`, so a machine with one and not the other skips instead of erroring (same class as the `cargo`-without-`clippy` fix on #734).

  5 behavioral tests still pass locally.

VytCepas:

- **Copilot — `bun pm ls` has no timeout while every other bun subprocess in the module does**: Correct, and the inconsistency is the tell. A hung `pm ls` would stall the entire pytest run rather than fail one test. Bounded at 600s, matching its siblings.

All 5 behavioral tests pass, full suite green, `ruff check` clean.

VytCepas:
**Review response (cycle 6):**

- **Copilot — the toolchain guard probes only `eslint-plugin-security`, so a partially-upgraded project still crashes**: Correct. A project that already has `eslint-plugin-security` but lacks `eslint-plugin-no-unsanitized` would skip `just setup` and then die at import time with exit 2 — the exact P1 this guard was added to prevent, surviving in a narrower form.

  The guard now probes **every** plugin `eslint.config.mjs` imports, and breaks out to `just setup` on the first miss.

  Verified against that precise state — `package.json` listing `eslint-plugin-security` and not `eslint-plugin-no-unsanitized`:

  ```
  security present, no-unsanitized missing -> guard fires: yes
  ```

  The contract test now asserts both plugin names appear in the guard, so adding a third import without extending the probe fails.

Full suite green, `ruff check` clean, rendered workflows actionlint-clean.

PR #734, opened 2026-07-10
fix(PI-725): make clippy see test code; give go and rust a typecheck recipe

VytCepas:
## The issue's framing was half wrong

Go and rust have no `just typecheck` recipe and no CI typecheck step — but they are **not untyped in CI**:

- `golangci-lint run` drives `go/analysis`, which type-checks.
- `cargo clippy` invokes the compiler front-end, which type-checks.

A type error already failed `just lint` for both. **No mypy-analog was missing.** Adding one would be ceremony.

## The real defect: clippy's scope

```
justfile.tmpl (before)
cargo clippy -- -D warnings -D clippy::pedantic -D clippy::cognitive_complexity -D missing_docs
```

No `--all-targets`, so clippy checks only lib/bin and **never sees `tests/`, `benches/` or `examples/`**. Verified against clippy 0.1.97 with a `tests/probe.rs` containing an outright type error (`let x: i32 = "not an integer";`):

| Command | Exit |
|---|---|
| `cargo clippy -- -D warnings …` (old recipe) | **0** — test code never checked |
| `cargo clippy --all-targets --all-features -- …` | **101** |

A Rust project could carry a test module that does not even type-check and pass its lint gate. `--all-features` closes the matching hole for code behind a non-default feature flag.

## Why the recipe is now two passes

Adding `--all-targets` naively **breaks clean projects**. `-D missing_docs` then applies to test targets and demands a crate-level `//!` in every `tests/*.rs`:

```
error: missing documentation for the crate
 --> tests/probe.rs:1:1
```

That would be a born-red scaffold — the #698 class of bug, again. So:

1. `cargo clippy --all-features -- … -D missing_docs` — public-API docs, lib/bin only.
2. `cargo clippy --all-targets --all-features -- …` — every target, no doc requirement.

Verified in all four directions: clean crate passes both; broken test fails pass 2; undocumented public fn fails pass 1; combining the two flags fails a clean crate.

## Uniform `just typecheck`

Added for go (`go vet ./...`) and rust (`cargo check --all-targets --all-features`), with the CI steps they lacked. Both verified to catch a type error in a **test** file and to pass once fixed. Redundant with lint by design — the point is that `just typecheck` means the same thing in all four languages, since that is the step CI and agents invoke.

## Verification — behavior, not text

`tests/integration/test_rust_lint_scope.py` runs real cargo. It includes `test_old_recipe_never_saw_a_broken_test`, which asserts the **old** command exits 0 — so if clippy ever changes and the bug becomes moot, the test says so instead of silently passing.

**Honest limitation:** it skips where cargo is absent, including this repo's CI. A skipped test is not a gate. #733 tracks the same problem for bun; the fix is the same shape (install the toolchain in CI).

Full suite green (2197 passed), `ruff check` clean, rendered workflows actionlint-clean.

Closes #725


VytCepas:

Both findings are correct, and both were verified against a real Go toolchain rather than reasoned about.

- **Codex P1 / Copilot — `just typecheck` runs before `Install just` in the Go job**: Fixing. Confirmed by reading the rendered workflow: `Typecheck` at line 77, `Install just` at line 82. On a fresh runner that is `just: command not found`, and it would have killed the job before tests or audit ran. The Go job is the only one where `Install just` sits *below* the lint steps (its lint comes from `golangci-lint-action`, not a justfile recipe), which is exactly why I dropped the step in the wrong place. Moved after `Install just`; the rendered order is now `Install just` (76) → `Typecheck` (83), pinned by `test_go_typecheck_step_comes_after_just_is_installed`.

- **Codex P2 — `go vet ./...` is red on a module with no sources**: Fixing. Verified:

  ```
  $ go vet ./...          # go.mod, zero .go files
  go: warning: "./..." matched no packages
  no packages to vet
  EXIT=1
  ```

  So a fresh Go scaffold would have been born red — the #698 class of bug, which this PR's own description warns about for rust and which I then reproduced in go. The sibling recipes (`test-cov`, `license`, `fuzz`) already guard with `find . -name "*.go" | grep -q .`; `typecheck` now uses the same idiom.

  Verified all three states against a real toolchain:

  | Module state | Exit |
  |---|---|
  | `go.mod`, no sources | **0** — "No Go sources yet — nothing to type-check." |
  | one clean source file | 0 |
  | source with a type error | **1** |

  The guard skips cleanly without weakening the gate.

Full suite green (2199 passed), `ruff check` clean, rendered workflows actionlint-clean.

VytCepas:

Both correct. Fixed.

- **Copilot — the `ci:` alias omits `typecheck` for go and rust**: Right, and it is the sharper form of what this PR is about. `just ci` is commented "what CI runs"; the moment CI gained a Typecheck step, the local gate diverged from it. Confirmed:

  ```
   86: ci: setup lint typecheck test-cov audit   # python
  180: ci: setup lint typecheck test audit       # node
  269: ci: lint test-cov audit                   # go    <- no typecheck
  374: ci: lint test-cov audit                   # rust  <- no typecheck
  ```

  Both now include it. `test_ci_alias_includes_typecheck` parametrizes over all four languages, so a future language cannot ship an alias that lies about what CI runs.

- **Copilot — the skip condition checks for `cargo`, but the tests need `cargo clippy`**: Right. clippy is a separate rustup component and is commonly absent on distro-packaged toolchains, so the module would have *failed* rather than skipped. The probe now runs `cargo clippy --version` and skips on a non-zero exit.

  Verified with a stub `cargo` that succeeds generally and exits 101 on `clippy`:

  ```
  SKIPPED [1] test_rust_lint_scope.py: no cargo+clippy (CI ships none)
  5 skipped
  ```

  And with the real toolchain, all 5 still run and pass.

VytCepas:

- **Copilot — `ruff check` will report `E302 expected 2 blank lines, found 1`**: **Not applying — false on both counts.**

  There *are* two blank lines before `_has_clippy`:

  ```
  27 | import pytest
  28 |
  29 |
  30 | def _has_clippy() -> bool:
  ```

  And E302 is not even active in this repo. Asking ruff for it explicitly:

  ```
  $ uv run ruff check --select E302 tests/integration/test_rust_lint_scope.py
  warning: Selection `E302` has no effect because preview is not enabled.
  All checks passed!
  ```

  `just lint` (which is `ruff check .`) has passed on every commit of this branch.

  Worth noting for the record: this module *did* have a real import-order violation (`I001`) earlier — a `git stash pop` restored a pre-`--fix` copy and I committed it. The thing that caught it was not a reviewer but `TestGitleaksPreCommitHook`, which runs the scaffolded pre-commit hook against this repo's staged snapshot. I initially misfiled that as a known environment flake; it failed deterministically on this branch and passed on `main`, which is what showed it was mine. Fixed and committed.

All checks green, no outstanding findings.

VytCepas:

All three correct. Fixed.

- **`_has_clippy()` probes without a timeout, during pytest collection**: Right, and collection is the worst place for it — a hung toolchain stalls the entire run rather than failing one test. Bounded at 60s, with `text=True`, and `TimeoutExpired`/`OSError` now mean "no clippy" (skip) rather than an error escaping collection.

- **Chained `split(...)` to extract the `typecheck` recipe raises `IndexError` instead of asserting** (×2): Right. A missing recipe is a *contract failure* and should read as one, not as a test error in a traceback. Adopted the `_recipe_body` regex helper from `tests/contracts/test_justfile.py:45` rather than inventing a second one.

  Verified the failure mode: extracting an absent recipe now raises `AssertionError: recipe 'typecheck' not found in the scaffolded justfile`, not `IndexError`.

Full suite green (2228 passed), `ruff check` clean; the 5 rust behavioral tests still pass with a real toolchain.

VytCepas:

- **Copilot — `next(...)` raises `StopIteration` if the `ci:` alias is missing**: Correct, and it is the same shape as the `_recipe_body` fix one cycle ago: a contract test should report a *contract failure*, not blow up with a traceback that says nothing about the contract.

  Now it collects the matching lines, asserts non-empty with the language named, then checks the alias. Verified: a justfile with no `ci:` alias produces `AssertionError: go: no `ci:` alias in the scaffolded justfile`, not `StopIteration`.

Full suite green (2228 passed), `ruff check` clean.

VytCepas:
**Review response (cycle 6):**

- **Copilot — these tests skip in CI because no Rust toolchain is provisioned, so real clippy regressions won't be caught there**: Correct, and already stated rather than hidden. The module docstring says it:

  > *Honest limitation: skips where cargo is absent — including this repo's CI, which installs no rust toolchain. A skipped test is not a gate; see #733.*

  The PR description says the same, and **[#733](https://github.com/VytCepas/project-init/issues/733)** exists to fix it: install bun *and* cargo in `lint-and-test`, then narrow or drop the skips so a missing toolchain **fails** in CI. Its acceptance criteria are the behavioral ones — reverting the `typescript@^5` pin, or a security rule from `error` to `warn`, must turn CI red.

  Not folding that into this PR. It touches CI provisioning for two toolchains and belongs with the identical bun gap from #729, not bolted onto a clippy-scope fix. Doing it here would mean this PR both introduces the tests and changes how CI installs Rust, with no separate place to verify the second half.

  Worth being precise about what is and isn't covered meanwhile: the *contract* tests (`--all-targets` present, `missing_docs` absent from that line, `ci:` alias includes `typecheck`, the go source-less guard) **do** run in CI. What skips is the behavioral proof that clippy actually misses `tests/` without the flag. So a regression in the recipe is caught; a regression in clippy itself is not, until #733.

All checks green, no outstanding code findings. Merging.

PR #748, opened 2026-07-11
chore(PI-637): Add dependency vulnerability scanning (pip-audit or Dependabot)

VytCepas:
## What

Adds dependency-vulnerability scanning to the repo's own CI, dogfooding what the scaffolded `ci.yml.tmpl` already ships.

- New `audit` justfile recipe: `uv run --with pip-audit pip-audit` (ephemeral, no dev-dependency to forget), wired into `just ci` and a discrete `Dependency vulnerability scan (pip-audit)` step in `lint-and-test`.
- The gate immediately caught real advisories in transitive deps pulled via `mkdocs-material` → `requests`: bumped `idna` 3.13→3.18 (PYSEC-2026-215) and `urllib3` 2.6.3→2.7.0 (PYSEC-2026-141/142) in `uv.lock`. Audit now reports **no known vulnerabilities**.

## Why

Gitleaks scans for committed secrets, but nothing checked for known CVEs in a correctly-spelled dependency already in the lockfile. The template ships this; the repo's own CI was behind its own product.

## Verification

- `just audit` → no known vulnerabilities.
- Full suite green: 2260 passed, 10 skipped.
- Broke it to prove the gate works: pre-bump, `pip-audit` reported 5 advisories and the step would have failed CI red; post-bump it passes.

Closes #637

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,14 @@ jobs:
         # a dip is visible in the log and never blocks an unrelated PR.
         run: uv run pytest --tb=short -q --cov --cov-report=term-missing
 
+      # Dependency vulnerability scan (#637). gitleaks (secret-scan job) catches
+      # committed secrets; nothing else here catches a correctly-spelled dep
+      # already in uv.lock with a known CVE. pip-audit is pulled in ephemerally
+      # via `uv run --with`, the same pattern the scaffolded ci.yml.tmpl uses —
+      # no dev-dependency to forget. Mirrors the template's `just audit` step.
+      - name: Dependency vulnerability scan (pip-audit)
+        run: just audit
+
   wheel-smoke:
     name: Wheel smoke test
     runs-on: ubuntu-24.04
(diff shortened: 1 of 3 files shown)


VytCepas:
**Review response:**
- [Codex P2: audit the docs extra before declaring the lockfile clean]: **Fixing — you're right, and the hole was live.** The recipe now runs `uv run --all-extras --with pip-audit pip-audit`, so every locked optional-dependency group (incl. `docs`) is in the audited environment. Verified the gap concretely: in a clean `uv sync --group dev` env, `urllib3`/`idna` aren't even installed (ModuleNotFoundError), so the default-env scan couldn't see them — it only caught them locally because my dev venv happened to have the docs extra present. With `--all-extras` the gate immediately surfaced a *second* docs-extra advisory the original scan missed: `pymdown-extensions` 10.21.2 (CVE-2026-46338), now bumped to 11.0.1 (docs still build under `mkdocs build --strict`). Thanks — this turned a gate-that-can't-fire into one that does.

PR #750, opened 2026-07-11
chore(PI-639): Confirm whether the no-mypy stance for this repo's own code is still intended

VytCepas:
## What

Adopts `mypy --strict` on `src/`, dogfooding the strict typecheck gate the scaffolder already ships (`templates/base/mypy.ini.tmpl` + `just typecheck` in the scaffolded `just ci`). Closes the "#639 — confirm the no-mypy stance" ticket by **reversing** it: the repo now runs the same gate as its product.

- **`mypy.ini`** — `strict = True`, `python_version = 3.11` (the requires-python floor / oldest matrix leg), scoped to `src/`.
- **`just typecheck`** recipe (mypy via `uv run --with`, no dev-dep to forget) → added to `just ci` + a CI `Typecheck` step in `lint-and-test` (so `ci-gate` covers it).
- **Annotated `src/` to zero strict errors** (was ~101 across 9 files). Behaviour-neutral: precise types where uniform, explicit `Any` only for heterogeneous parsed-TOML/JSON blobs, **zero `# type: ignore`**.
- **Docs stance flipped**: CLAUDE.md + copilot-instructions no longer say "no mypy"; also fixed the stale CLAUDE.md line claiming the repo runs no pip-audit (it has since #637).

## Non-annotation touches (each commented + verified inert)

- two `or (0,0,0)` narrowings on already-guaranteed-parseable versions;
- one `assert conflicts is not None` the surrounding `protect = conflicts is not None and …` already guarantees (can't fire);
- one `str(raw)` no-op (raw already matched a str tier);
- one local `text`→`decoded` rename (shadowing fix, behaviour identical).

## Verification

- `just typecheck` → "Success: no issues found in 11 source files".
- ruff clean; full suite **2262 passed, 10 skipped** — unchanged count, i.e. no behaviour change.
- Non-annotation hunks reviewed individually against surrounding invariants.

Closes #639


VytCepas:
**Review response:**
- [Codex: `or (0,0,0)` sort fallback silently orders unparseable versions as 0.0.0, hiding an invariant break]: **Fixing — applied to both occurrences.** In `migration_notes.notes_for_span` I now carry the already-parsed tuple `v` (from the comprehension's walrus) through to the sort, so it neither re-parses nor needs any fallback — an unparseable version can't reach the sort because the comprehension already dropped it. In `upgrade.py` (same pattern, single value) I replaced the fallback with an explicit `assert parsed is not None` so the 'notes non-empty ⇒ current parseable' invariant fails loudly instead of formatting `v0.0.0`. Both are behaviour-neutral (the invariants already held); suite still 2262 passed.

VytCepas:
**Review response (cycle 1 — Copilot):**
- [migration_notes.py `or (0,0,0)` sort fallback]: **Already fixed** in the prior cycle-1 commit (before this review ran) — the sort now carries the parsed tuple `v` from the comprehension, so there is no re-parse and no fallback at all. Stale comment; resolving.
- [upgrade.py:458 `_record_fields` returns dict[str,str] without validating json values]: **Fixing — good catch.** `json.loads` is typed `Any`, so mypy accepted the `dict[str, str]` annotation while runtime could carry non-str values from a hand-edited config. Added `_str_map`: a record field is accepted only when it is a flat str->str mapping, else it returns None so `_parse_record_block` raises `UpgradeError` (fail loud — a record mustn't silently drop a variable, unlike `read_base` which filters the merge base). The annotation is now a validated invariant.
- [mypy.ini:4 'the same way ruff exempts them']: **Fixing.** Reworded to state plainly that mypy is invoked on `src/` only; dropped the inaccurate ruff-exemption comparison (ruff uses per-file rule ignores, not a blanket exemption).

ruff + mypy --strict clean; suite 2262 passed.

PR #752, opened 2026-07-11
docs(PI-684): Worked grounding example for the diagram skill

VytCepas:
## What

Three related polish items on the templated `diagram` skill (WS4/WS6/WS2 of epic #677), edited in the `templates/fallback` source and propagated to the 4 per-surface copies + the plugin via `sync_plugin.py`.

- **#684 (WS6)** — worked grounding example + the **labels-carry-paths** rule: file-backed nodes carry their repo-relative path (e.g. `(src/project_init/scaffold.py)`), so a later check can flag a diagram pointing at a moved/deleted file. This is what makes #688's dangling-path lint effective on diagrams. Also confirmed the `CODE_MAP.md (if present)` reference degrades for non-Python stacks.
- **#682 (WS4)** — a **"Not for data charts"** routing note near the top: structural diagrams only; data/number charts go to a stack-appropriate plotting library, since scaffolded projects have no dataviz skill. **Decision recorded on #677**: don't template dataviz for now.
- **#680 (WS2)** — the **commit-not-gitignore** policy for rendered assets, stated in the skill's Finalize section, plus `docs/diagrams/**/*.svg linguist-generated=true` in the `.gitattributes` template (collapse render diffs; `.mmd` stays the diffable artifact). No wizard flag (a convention, not a selectable concern).

## Tests

- New exact-case assertions in `test_diagram_skill.py` for all three notes.
- Re-pinned **only** the `.gitattributes` hash across the 8 byte-identity baselines — verified no other key drifted (the diagram `SKILL.md` is already in `_ADDED_SINCE_BASELINE`).
- Full suite: 2262 passed, 10 skipped.

Closes #684
Closes #682
Closes #680

### DIFF
--- a/plugins/project-init-workflow/skills/diagram/SKILL.md
+++ b/plugins/project-init-workflow/skills/diagram/SKILL.md
@@ -106,13 +106,13 @@ source is the artifact.
 - Commit the whole `<slug>/` folder: source **and** rendered picture
   together. The source is the artifact of record for diffing/regeneration;
   the picture is what a human opens without tooling — both ship, always.
-  **Commit the render, don't gitignore it:** GitHub renders ```` ```mermaid ````
+  **Commit the render, don't gitignore it:** GitHub renders ```` ```mermaid ``` ````
   fences but not `.mmd` files as images, so the checked-in picture is the only
   way a human sees the diagram on the forge without local tooling. It's marked
   `linguist-generated` in `.gitattributes` so it doesn't inflate diffs or
   language stats, and the always-re-render rule (§3) keeps it from going stale.
 - Embedding:
-  - GitHub/GitLab render ```` ```mermaid ```` fences in markdown natively —
+  - GitHub/GitLab render ```` ```mermaid ``` ```` fences in markdown natively —
     inline the source in docs/README where useful.
   - mkdocs needs one-time config; offer to add it:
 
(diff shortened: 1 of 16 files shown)


VytCepas:
**Review response:**
- [Copilot: inline mermaid-fence illustration shows only the opening fence]: **Fixing.** Both inline illustrations now show the full ```mermaid ... ``` open+close pair inside the four-backtick wrapper. The actual fenced code block (the worked grounding diagram) already had a proper open+close and is unchanged.

VytCepas:
**Review response (cycle 1 — Copilot):**
- [vault-based rendered SVGs won't be collapsed as generated]: **Partly a premise correction, and reworded for accuracy.** Vault renders *are* already collapsed — the base `.gitattributes` carries `.agents/vault/** linguist-vendored=true` (line 17), and `linguist-vendored` collapses diffs and drops language stats the same way `linguist-generated` does. So behaviour is already consistent regardless of where the diagram folder lives. The real issue was my skill *wording*, which said the render is marked `linguist-generated` (accurate only for `docs/diagrams/`). Reworded to name both: `docs/diagrams/**/*.svg` → generated, vault → already vendored. I did **not** add a vault `linguist-generated` line — it would contradict the existing vendored rule on the same path.

PR #753, opened 2026-07-11
docs(PI-683): Align ADR template and Obsidian design vault with diagram folder convention

VytCepas:
## What

WS5 of epic #677 — align two design-artifact docs with the folder-per-diagram convention so nothing contradicts the diagram skill.

- **ADR template** (`templates/base/.../adr-template.md`): added author guidance in the leading (deleted-before-commit) comment — draw ADR diagrams with the diagram skill and **link** the `docs/diagrams/<slug>/` (or `.agents/vault/design/<slug>/`) folder rather than pasting a render, so the ADR doesn't drift when the diagram is re-rendered (map-not-territory).
- **Vault `design/` README** (`templates/obsidian/.../design/README.md`): states diagrams get a folder each holding the `.mmd` source **and** rendered picture, re-rendered on change — matching the diagram skill's vault path + always-render rule.

## Tests

- New assertions: `TestAdrToolchain.test_madr_template_links_diagrams_not_pastes_them` + `test_design_readme_states_folder_per_diagram_convention`.
- Re-pinned only the two changed hashes across the 8 byte-identity baselines (verified no other key drifted).
- Full suite: 2264 passed, 10 skipped.

Closes #683


VytCepas:
**Review response:**
- [Codex: linking docs/diagrams/<slug>/ from an ADR at .agents/docs/adr/ yields a broken relative link]: **Fixing — correct catch.** Reworded the guidance to (a) name the paths explicitly as *repo-root* paths and (b) show the correct href written relative to the ADR's home: `../../../docs/diagrams/<slug>/` (and `../../vault/design/<slug>/` for the vault). Verified the level count from `.agents/docs/adr/`.

VytCepas:
**Review response:**
- [Copilot: the test would still pass if the template dropped the anti-paste warning or the vault path]: **Fixing — right, it wasn't guarding its own intent.** Added assertions for `vault/design/<slug>/` and `don't paste` alongside the existing docs-path + `link that folder` checks, so removing any of the three load-bearing parts now fails the test.

VytCepas:
**Review response (cycle 2 — Copilot):**
- [docstring says 'three' but asserts four]: **Fixed** — reworded the docstring to describe the intent without a stale count.
- [vault path matched a loose substring; assert the example hrefs]: **Fixed** — the test now asserts the exact ADR-relative hrefs `../../../docs/diagrams/<slug>/` and `../../vault/design/<slug>/`, so a regression in the worked link form (not just the folder name) fails. Kept the anti-paste + link-that-folder asserts.

VytCepas:
**Review response (cycle 3 — Copilot):**
- [README says generic 'picture' but the skill standardizes on .svg]: **Fixed** — the vault README now names the rendered `.svg` (via mermaid-cli), matching the skill; `.mmd` source vs `.svg` render is explicit.
- [README test assertions too weak]: **Fixed** — the test now asserts the folder-per-diagram rule, the `.mmd`/`.svg` pair, `re-rendered on every source change`, and `both are committed`, on whitespace-normalized text so a line wrap can't let a phrase slip. Dropping any load-bearing part now fails.

VytCepas:
**Review response (cycle 4 — Copilot):**
- ["via mermaid-cli" is a volatile implementation detail]: **Fixing — and you're right by our own rule.** Removed the tool name; map-not-territory (the very principle #677 codifies) says the README shouldn't restate the render tool when the `diagram` skill already owns the command. Kept the stable `.svg` artifact convention and pointed at the skill for the command.
- [README test assertions too weak]: **Already addressed** — this looks like a re-post of the cycle-3 comment; the test was strengthened in the previous commit and now asserts `folder each`, ``.mmd` source`, ``.svg``, `re-rendered on every source change`, and `both are committed` on whitespace-normalized text. Resolving as stale.

PR #754, opened 2026-07-11
chore(PI-681): Drift convention/helper for personal ~/.claude skills vs templates

VytCepas:
## What

WS3 of epic #677 — give the one un-guarded skill-sync edge a drift signal: the maintainer's personal `~/.claude/skills/<name>/` copies (outside VCS) vs their `templates/` source. This is the edge that let the `diagram` skill silently diverge and motivated the epic.

- **`tools/skill_drift.py`** — stdlib, **advisory** (always exits 0, since the personal dir isn't versioned here). Diffs each declared shared skill's personal copy against its `templates/fallback` source and prints the drift. `SHARED_SKILLS = {diagram}` for now, extensible.
- **`just skill-drift`** recipe.
- **CLAUDE.md** — a **bidirectional** convention: an edit to a shared skill's personal copy must be ported to its template source the same session (then `sync_plugin.py` + `sync-claude`), and a template edit reflected back.
- **`tests/contracts/test_skill_drift.py`** — guards that every `SHARED_SKILLS` source exists (a typo'd path would make the tool report clean for a skill it never compared) and exercises the in-sync / drift / missing verdicts against a fake personal dir.

## Verification

- `just skill-drift` correctly flags the current real drift (personal diagram skill is behind the template) and exits 0.
- ruff + `mypy --strict` clean; full suite 2268 passed, 10 skipped (+4 new).

Closes #681


VytCepas:
**Review response:**
- [Copilot: the contract test is vacuously true if SHARED_SKILLS is emptied]: **Fixing — correct, that's the repo's own 'a test that can't fail' trap.** Added `assert skill_drift.SHARED_SKILLS` before the loop, so an emptied map (which would make `just skill-drift` a silent no-op) fails CI.

PR #755, opened 2026-07-11
chore(PI-688): Map-not-territory rule + deterministic prose/diagram staleness checks

VytCepas:
## What

WS10 of epic #677 — codify **map-not-territory** and add deterministic guards for hand-maintained facts that had no check tying them to their source.

1. **Codify map-not-territory** — new section in the memory `SCHEMA.md` (a memory may assert only mechanically-verifiable facts — repo paths, boundaries, ownership — or immune-to-change ones — rationale, decisions; volatile specifics like tool names / thresholds / pins / line numbers are *pointed at, not restated*). The diagram skill already states it (PI-684); locked here with a test assertion.
2. **Strip the coverage-floor value (`70`)** from the four rule docs (`python/go/rust/typescript.md`) — the value is the `justfile`/`bunfig.toml`'s to own; docs now say "threshold per justfile". (No test asserted the value; the `#688` audit noted `test_justfile.py` checks only the flag name.)
3. **DAG contract test** — the reminder's ASCII lifecycle diagram must name exactly `dag_workflow.py`'s `GRAPH` node set; fails on a GRAPH node added without updating the reminder.
4. **Committed-diagram path lint** — every repo-relative path a `.mmd` names must resolve, so a diagram pointing at a moved/deleted file fails CI (makes diagrams safe as navigational memory).

## Scope decision (documented in the test)

The dangling-path lint runs on **`.mmd` diagrams only**, not `docs/**/*.md` or skills. I scanned first: those corpora describe the *scaffolded output* (`.agents/memory/MEMORY.md`) and template-relative names (`dot_agents/hooks/...`), so a repo-root existence check false-positives on ~49 legitimate files — a noise machine, not a guard. Diagrams reference real repo paths, so the check is precise there. This is the issue's "warn or fail — decide and document": **fail**, on `.mmd` only.

## Verification

- Both guards pass; break-it confirmed (a fake missing `.mmd` path is flagged; the DAG test compares exact node sets).
- Re-pinned only the `SCHEMA.md` hash across byte-identity baselines (rule docs are already excluded).
- ruff + `mypy --strict` clean; full suite 2270 passed, 10 skipped.

Closes #688

### DIFF
--- a/templates/auto/dot_agents/memory/SCHEMA.md
+++ b/templates/auto/dot_agents/memory/SCHEMA.md
@@ -49,6 +49,18 @@ One fact per file — move longer material to the vault.
 - Anything already in `CLAUDE.md` or `project-init.md`
 - Large documents (put those in `vault/`)
 
+## Map, not territory
+
+A memory — like any cached artifact (a diagram, a doc) — may assert only facts
+that are **mechanically verifiable** (repo-relative paths, module boundaries,
+ownership) or **immune to change** (a rationale, a decision and its date).
+Volatile specifics — tool names, thresholds, version pins, line numbers — must
+be **pointed at, not restated**: name the file that owns the value, don't copy
+the value in. Restated "territory" rots silently the moment its source moves,
+and no gate catches it: `lint_memory.sh` can flag a dangling backtick-`path.ext`
+but not a threshold that has drifted from the `justfile`. Cheap-to-regenerate
+artifacts get regenerated, not cached; underivable rationale goes here or to an ADR.
+
 ## Relationship to vault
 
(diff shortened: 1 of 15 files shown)


VytCepas:
**Review response:**
- [Copilot + Codex P2: rust rule's bare `cargo llvm-cov` doesn't enforce the threshold]: **Fixing — you're both right.** Changed it to `just test-cov`, matching the Go/Python rules. Coverage is the one line where the raw tool ≠ the recipe: `cargo llvm-cov` alone reports but doesn't gate (the recipe adds `--fail-under-lines`), so pointing at the recipe is both correct and map-not-territory (the threshold stays the justfile's).
- [Copilot: DAG-drift message only prints the ASCII set]: **Fixing** — the assertion now prints the symmetric difference (only-in-GRAPH / only-in-ASCII) so a drift is fixable at a glance.

PR #756, opened 2026-07-11
feat(PI-747): verify-test-strength skill: agentic mutation loop to prove guards can fail

VytCepas:
## What

A new default-on, user-invocable **`verify-test-strength`** skill: the automated form of the break-it discipline. Given a load-bearing test, it runs the project's mutation tool, surfaces surviving mutants, and drives the agent to strengthen tests until they're killed — because coverage is cheap and lies (an LLM suite can hit 100% coverage at ~4% mutation score).

Motivated by #743/#745 (the *instruction-level* floor) + the literature: the measured gain comes from the **iterative mutation loop**, not a one-off "try to break it". This session is the same finding lived out — the vacuity bug recurred 5× and was caught only by the manual loop.

## The skill

- Scope to one module + its test → run mutation tool → kill survivors one at a time (assert on the behaviour the mutant changed) → re-run (~4 rounds).
- **Python-gated** to `mutmut` / `just test-mutation` (the only wired runner). Names StrykerJS / cargo-mutants / gremlins but flags them **not-scaffolded** and drops to a manual break-it loop rather than faking a score.
- **Honest reporting**: equivalent mutants, dead code (a finding, not a forced test), genuinely-untestable branches — reported, never forced green.
- **Selective** by design (the loop is ~32% more tokens) — load-bearing guards, not the whole suite.

## Decisions (from #747's "decision needed" section)

Python-only PoC now · dedicated skill (not folded into `/verify`) · manual/selective invocation. Per-language mutation recipes (Node/Rust/Go) remain a separate prerequisite for cross-language coverage — the skill is language-gated to where a runner is actually wired.

## Wiring

Auto-synced to the plugin + 4 per-surface copies (`sync_plugin.py` globs skills). Registered in INDEX / README / project-init tables; plugin `0.6.0 → 0.7.0`; new SKILL.md added to both byte-identity `_ADDED_SINCE_BASELINE` sets.

## Verification

- New content test (present + default-on, method coverage, listed in the 3 tables).
- ruff + `mypy --strict` clean; full suite **2273 passed**, 10 skipped.

Closes #747

### DIFF
--- a/plugins/project-init-workflow/.agents-plugin/plugin.json
+++ b/plugins/project-init-workflow/.agents-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "project-init-workflow",
   "displayName": "project-init Workflow",
-  "version": "0.6.0",
+  "version": "0.7.0",
   "description": "Forge-agnostic quality and safety hooks plus general skills from project-init: commit gating, lint-on-edit, prod-safety guard, and session bootstrap. The always-on core payload; GitHub lifecycle automation lives in the separate project-init-lifecycle plugin.",
   "author": {
     "name": "Vytautas \u010cepas",
--- a/plugins/project-init-workflow/skills/verify-test-strength/SKILL.md
+++ b/plugins/project-init-workflow/skills/verify-test-strength/SKILL.md
@@ -1,0 +1,91 @@
+---
+name: verify-test-strength
+description: Prove a load-bearing test can actually fail — run mutation testing on a target module, then iteratively strengthen its tests until surviving mutants are killed. The automated form of the break-it discipline, for guards that must not silently pass.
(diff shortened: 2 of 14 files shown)


VytCepas:
**Review response:**
- [Codex P2: scoping `[tool.mutmut] source_paths` in pyproject narrows the nightly CI mutation job]: **Fixing — real footgun, verified.** The nightly `mutation-tests` job reads `source_paths` from `pyproject.toml` (ci.yml.tmpl), so a committed narrow scope shrinks CI coverage to one module. Reworded §2 to target the run without touching the committed config: `mutmut run <path>` per-run filter, or temporarily narrow-and-restore `source_paths` before committing — never commit the narrow scope. `just test-mutation` still means the full CI scope.

PR #758, opened 2026-07-11
chore(PI-757): Quality-gate policy doc + required-gate drift guard (epic #751)

VytCepas:
## Summary

Epic #751 foundation — the canonical, deliberately-calibrated quality/security-gate policy, plus a drift guard.

- **`docs/development/quality-gates.md`** — one policy for both surfaces: the scaffolded template is canonical, this repo's own CI is a faithful *subset*. Full inventory table with a **keep/promote/demote/drop** decision + rationale per gate; the four-gaps decisions; a "deliberately not adopted by the repo" section (scorecard/fuzz/mutation/container). Declares the repo's required-check set in a machine-readable marker.
- **`tests/contracts/test_quality_gate_policy.py`** — fails CI if `ci-gate.needs` in `ci.yml` drifts from the documented required set (and asserts CLAUDE.md links the doc).
- **`CLAUDE.md`** — links the policy from the Reference (on demand) section.

## Premise correction

The epic assumed the repo lacks a `ci-gate` aggregator. It has one. The real gap is the repo lagging the template on coverage floor / `ruff format --check` / semgrep / license-scan — addressed in follow-up PRs under #751 (foundation-first).

## Verification

- Break-it check: desyncing the required-gates marker fails the drift guard; restoring passes. ✅
- Full suite: 2277 passed, 10 skipped. `ruff check .` clean. `mkdocs build --strict` clean.

Closes #757

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -88,3 +88,10 @@ CI optimization strategies, the extend-the-infrastructure table (skill vs
 hook vs agent spec), and this repo's non-goals live in
 [`docs/development/repo-reference.md`](docs/development/repo-reference.md)
 — consult when adding capabilities or tuning CI, not needed per turn.
+
+The canonical quality/security-gate policy — every gate the repo and the
+scaffolded template run, each with a keep/promote/demote/drop decision, and the
+repo's single required-check set — lives in
+[`docs/development/quality-gates.md`](docs/development/quality-gates.md)
+(epic #751). Update it in the same PR when you change a gate; a contract test
+fails if the repo's required checks drift from it.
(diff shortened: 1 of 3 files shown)


VytCepas:
- `test_quality_gate_policy.py:15` (misleading pyyaml comment): Fixing — the repo does use pyyaml in tests, so "no pyyaml, per repo convention" was wrong. Reworded to the actual reason: the marker isn't YAML and `ci-gate.needs` is a single well-known line, so a full YAML parse buys nothing here.
- `test_quality_gate_policy.py:44` (set drops duplicates + unpinned encoding): Fixing both — added a `_split_gates` helper that asserts on a duplicated gate before returning the set (a doubled gate now fails loudly instead of silently passing), and pinned `encoding="utf-8"` on all three reads.

PR #760, opened 2026-07-11
feat(PI-759): lighter pre-push gate (both surfaces); CI parallelization deferred to #762

VytCepas:
## Summary

Reduces the local push cost now; defers the CI wall-clock win because parallelizing surfaced a real (rare) test interference.

### Ships: lighter pre-push (both surfaces) — deterministic
The pre-push hook re-ran the **full** `just ci` before every push, duplicating what CI is about to do. New **`fast-ci: lint test`** recipe (no typecheck/coverage/audit) becomes the pre-push gate in both `.githooks/pre-push` (repo) and `pre-push.tmpl` (scaffolded output). CI stays the full backstop → fast push→CI loop, nothing dropped. Security stays cheap-and-local (gitleaks pre-commit untouched).

### Deferred: CI pytest parallelization → #762
`-n auto` **proved the win** on this PR's first CI run — the Tests step dropped **5m08s → 2m06s**. But that same run **failed once**: `test_lint_passes_on_clean_scaffold` read a wrong/corrupt tree under CI's worker scheduling — the exact cross-test interference the original serial comment named (this suite has known coupling; see the `_frozen_templates` session fixture). Local `-n auto` and `-n 4` did **not** reproduce it, so it can't be root-caused/isolated safely here. Shipping nondeterministic CI would cause flaky reds on unrelated PRs — the opposite of faster time-to-merge. CI reverts to **serial**; `just test` keeps `-n auto` for local dev speed. Parallelizing CI safely (root-cause, or split into parallel *jobs*) is #762.

## Tests
- New parametrized `test_fast_check_is_lighter_than_ci` (python/node/go/rust).
- Tightened `test_git_pre_push_runs_fast_ci` to assert the real `just fast-check` invocation (the old `just ci` substring assertion now only matches a comment — it would have passed for the wrong reason).
- shellcheck clean on both hooks.

## Scope
Epic #751. Matrix→nightly (cost) = #761. Safe CI parallelization = #762.

Closes #759

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,30 +1,31 @@
 #!/usr/bin/env bash
 # pre-push gate for the project-init repo itself (dogfood).
 #
-# Runs the SAME `just ci` (lint + tests) that CI runs, before the push leaves
-# the machine — so nothing red reaches a PR and the push→CI-fail→fix→re-push
-# loop is caught locally. This is the local↔CI parity gate whose absence let a
-# fresh-scaffold lint failure ship (the scaffolder's CI caught it only after a
-# push, and even then incompletely).
+# Runs the fast local gate `just fast-check` (lint + parallel tests) before the
+# push leaves the machine — so the common break (a failing test or lint error)
+# is caught locally instead of via the push→CI-fail→fix→re-push loop. It is
+# deliberately lighter than the full `just ci`: typecheck + audit run in CI,
+# which stays the hard backstop, so we don't re-run the whole gate on every push
+# (PI-759).
 #
 # Enabled automatically by `just setup` (git config core.hooksPath .githooks).
 # Bypass in an emergency with: git push --no-verify
 set -euo pipefail
 
 if command -v just >/dev/null 2>&1; then
-  # `just ci` tests the working tree, but the push ships the committed tree.
-  # Skip (fail-open) on a dirty worktree rather than test a tree that isn't what
-  # gets pushed — a WIP fix could mask a failure, or unrelated WIP fail a clean
-  # push. CI stays the backstop.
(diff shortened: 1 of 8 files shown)


VytCepas:
**Review response (Copilot) — all three applied:**
- `test_justfile.py:108` (header-only guard misses a heavy body): Fixed — added `assert _recipe_body(text, 'fast-check').strip() == ''` so `fast-check` must stay dependency-only; a stray `just typecheck`/`audit` body line now fails even if the header still reads `lint test`.
- `pre-push.tmpl:66` (overclaimed fail-open): Fixed — reworded to 'fail-open when `just` or the `fast-check` recipe is absent', and noted that with both present the recipe itself decides (day-one guards skip; a genuinely missing linter fails closed).
- `.githooks/pre-push:18` (repo hook diverges from template): Fixed — added the same `just --show fast-check` probe so the repo hook fail-opens (not fail-closes) when the recipe is unavailable, matching the template and the repo↔template consistency rule.

VytCepas:
- `justfile.tmpl` node `fast-check` recipe vs the `fast-check` property-testing library: Fixed by **renaming the recipe** `fast-check` → `fast-ci` across all four languages, both pre-push hooks (repo + template), and the tests. Chose rename over a clarifying comment because `fast-check` is *the* JS property-testing library — a Node dev typing `just fast-check` would reasonably expect property tests, not lint+tests. `fast-ci` also reads as 'the fast subset of `ci`'. The library references (`bun add -d fast-check`, the fuzz-section comments) are untouched, so the collision is gone.

VytCepas:
**Review response:** aligned the PR description to the actual recipe name `fast-ci` (the code, hooks, tests, and docs all use `fast-ci`; `fast-check` remains only as the JS property-testing library). No code change.

PR #763, opened 2026-07-11
chore(PI-589): CI clustering policy + concurrency-cancel

### LINKED ISSUE #589: Consolidate CI checks: one required gate + category grouping instead of a flat list of separate checks
## Problem

A PR currently shows ~12 separate check rows sourced from 7 workflow files:

- `ci.yml` — lint-and-test, wheel-smoke, secret-scan, shellcheck, changes, macos-portability, windows-portability
- `validate-pr.yml` — pr-title, closes-keyword
- `board-automation.yml` — board-move
- `review-status.yml` — set-review-status
- external Copilot reviewer

This is a lot of separate, uncategorized checks to scan. Goal: reduce the cognitive load / visual noise without losing signal.

## Key distinction (design guidance)

"Reduce the gates" bundles two things that pull in opposite directions:

- **Required status checks** (what actually blocks a merge) — *fewer is strictly better*.
- **Jobs** (the rows you see running) — *collapsing into one mega-job is usually a mistake*: it loses parallelism (jobs run concurrently on separate runners), loses failure isolation ("Security failed" vs "gitleaks failed at line X"), and erases the deliberate blocking-vs-advisory split (semgrep/scorecard/license/fuzz are intentionally non-blocking) and scheduled-only triggers.

Target model: **one thing that gates, many things that inform.**

## Notable findings

1. The repo's own `ci.yml` has **no aggregator job**, yet the scaffolded template (`ci.yml.tmpl`) it ships *does* — `ci-gate`, documented as "the SINGLE required status check for branch protection." The product teaches the pattern the repo itself doesn't use. That's the real inconsistency.
2. The security scans are heterogeneous (semgrep=Python, trivy=Docker, gitleaks=full-history checkout, scorecard=scheduled+OIDC, license=per-language) with different triggers and blocking status — they genuinely shouldn't share one runner.

### Scope (which CI)
- **(a) Repo's own CI** — project-init's PR checks only. Lowest risk; doesn't change what scaffolded projects receive.
- **(b) Scaffolded template** (`ci.yml.tmpl`) — the product every generated project inherits. Higher stakes; the granular gates are part of its value, so prefer "one required gate + grouping" over deleting scans.
- **(c) Both** — same pattern applied consistently.

### Depth (how aggressive)
- **(a) Gate + UI grouping only (recommended)** — add a single required `ci-gate` aggregator that `needs:` the blocking jobs; set branch protection to require only that context. Regroup existing jobs into ~3 named workflow files (**CI / Security / PR checks**) so GitHub's checks UI shows 3 headings instead of a flat list (GitHub groups checks by workflow). No jobs removed; no parallelism or failure-isolation lost.
- **(b) Above + fold trivial jobs** — also merge the instant PR-metadata checks (pr-title, closes-keyword, board-move) into one "PR hygiene" job.
- **(c) Merge scans into category jobs** — collapse security scans into a single "Security" job and quality into one "Quality" job. Fewest rows, but slower and loses per-scan failure isolation and the blocking/advisory split. **Advised against.**

## Recommendation

Start with **Scope (a) repo-only** at **Depth (a) gate + grouping** (optionally (b)). Net effect: **1 required check, ~3 tidy UI groups**, zero loss of signal or speed. If it proves out, replicate the same pattern to the template (Scope b/c).

## Acceptance criteria

- [ ] Repo `ci.yml` gains a single `ci-gate` aggregator job (`needs:` the blocking jobs, `if: always()`, fails unless all succeeded) — mirrors the template's `ci-gate`.
- [ ] Branch protection updated to require **only** the `ci-gate` context (plus whatever review/lifecycle contexts are intentionally required); document the change.
- [ ] Existing jobs regrouped into ~3 named workflow files so the checks UI shows CI / Security / PR-checks headings.
- [ ] Non-blocking/advisory scans stay non-blocking (not in `ci-gate`'s `needs`); scheduled-only jobs stay scheduled.
- [ ] No loss of parallelism (jobs stay separate; only the *required contexts* collapse).
- [ ] (If Scope b/c) same one-gate + grouping pattern applied to `ci.yml.tmpl`, with contract tests updated.

### Priority
low

### Notes
Spun out of a discussion during the #574–#580 security-gate work. Do the branch-protection change carefully — requiring only `ci-gate` is robust to matrix renames, but the required-context list must be updated in the same change or merges will block on a now-missing context.

VytCepas:
## Summary

Consolidates the CI check surface **deliberately** — the honest conclusion of #589 is that the high-value cluster already exists, and the tempting ones are traps.

- **`quality-gates.md` — new "CI clustering policy" section:**
  - **DO** one required check (`ci-gate`) — already in place; this is the whole of "reduce what blocks a merge."
  - **DON'T** merge jobs into a mega-job — kills parallelism (wall-clock = sum not max), can't span OSes, loses failure isolation + conditional/scheduled triggers + per-job re-run.
  - **DON'T** merge workflow files — doesn't reduce rows, widens permission scope (board/review need write), enlarges YAML blast radius (#719).
  - Composite "setup" action: **considered, not adopted** — only `Install uv` repeats (5×) and renovate keeps those pins in sync, so it's indirection for negligible gain.
- **`ci.yml` — `concurrency: cancel-in-progress`:** a new push cancels the superseded run, saving runner-minutes on review-cycle churn (the one low-risk consolidation win).

## Why not more
Mega-job / file-merge consolidation trades away parallelism, isolation, and least-privilege for cosmetics — and mega-job makes CI *slower*, against the epic's time-to-merge goal. Documented so nobody "tidies" the parallelism away later; the drift guard from #758 keeps the required-check set honest.

## Verification
mkdocs `--strict`, workflow-lint gate, drift guard, full suite (2281 passed).

Closes #589

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,9 +12,10 @@ permissions:
 # Cancel a superseded run when a new commit is pushed to the SAME pull request
 # (PI-589). Review cycles re-push often; finishing an obsolete run just burns
 # runner-minutes. Gated to pull_request events so a `main` push is never
-# cancelled — same schedule-safe form the template ships.
+# cancelled. `github.event_name` is in the group key so push and PR runs never
+# share a group — same schedule-safe form the template ships.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,13 +9,13 @@ on:
 permissions:
   contents: read
 
-# Cancel a superseded run when a new commit is pushed to the same ref (PI-589).
-# Review cycles re-push often; without this, every push leaves its predecessor
-# running to completion, burning runner-minutes on results nobody will read.
-# Keyed by workflow + ref; `main` pushes are not grouped with any PR's ref.
+# Cancel a superseded run when a new commit is pushed to the SAME pull request
+# (PI-589). Review cycles re-push often; finishing an obsolete run just burns
+# runner-minutes. Gated to pull_request events so a `main` push is never
+# cancelled — same schedule-safe form the template ships.
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   lint-and-test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,14 @@ on:
 permissions:
   contents: read
 
+# Cancel a superseded run when a new commit is pushed to the same ref (PI-589).
+# Review cycles re-push often; without this, every push leaves its predecessor
+# running to completion, burning runner-minutes on results nobody will read.
+# Keyed by workflow + ref; `main` pushes are not grouped with any PR's ref.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   lint-and-test:
     name: Lint and test
--- a/docs/development/quality-gates.md
+++ b/docs/development/quality-gates.md
@@ -63,6 +63,14 @@ render separately) and it widens the permission scope (board/review jobs need
 `projects`/`statuses` write), entangles triggers, and enlarges the blast radius
 of a single YAML error (#719).
 
+**Supersedes the original #589 proposal.** #589 first suggested regrouping jobs
+into ~3 workflow files to get GitHub checks-UI headings. That's the "merge the
+workflow files" move above — investigating it showed it doesn't reduce the visible
+rows and costs permission scope + trigger clarity, so we deliberately *don't* do
+it. Closing #589 with this policy is the reasoned resolution, not a dropped
+requirement: the cognitive-load goal is met by the one-required-gate model plus
+the drift guard, without the file reshuffle.
+
 **Adopted low-risk wins (both surfaces):** `concurrency: cancel-in-progress` on
 `ci.yml` — a new PR push cancels the superseded run, saving runner-minutes on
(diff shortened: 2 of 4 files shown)


VytCepas:
**Review response (Codex, P2 — good catch):** Applied to the template. `templates/base/dot_github/workflows/ci.yml.tmpl` now ships the same concurrency block, and since the template has nightly/weekly crons I made both surfaces **schedule-safe**: `cancel-in-progress: ${{ github.event_name == 'pull_request' }}` — only superseded PR runs are cancelled; scheduled (fuzz/mutation/scorecard) and base-branch-push runs always complete. Repo ci.yml tightened to the same form (protects main-push runs too). Doc entry updated to 'both surfaces', and added `test_concurrency_cancels_only_pull_requests` asserting the PR-gated form (a bare `cancel-in-progress: true` now fails the test).

VytCepas:
- `test_ci_runs_on.py` (brittle substring): Fixed — now parses via `tests/workflow.load_workflow` and asserts `concurrency.cancel-in-progress` structurally (`is not True` catches a bare-true that would kill cron runs; exact-match confirms the gated expression). Verified the check fails on a bare `true`.
- `quality-gates.md` (#589 supersession): Added an explicit note that this policy supersedes #589's original 'regroup into ~3 workflow files' idea — investigated, it doesn't reduce the checks-UI rows and costs permission scope/trigger clarity, so closing #589 with the one-gate + drift-guard model is the reasoned resolution, not a dropped requirement.

VytCepas:
**Review response (Copilot) — both applied:**
- Concurrency group now includes `github.event_name`, so scheduled / base-branch-push / PR runs get distinct groups and never serialize behind each other (cancel-in-progress stays pull_request-gated). Applied to repo + template.
- Test asserts `cancel-in-progress` is present before indexing → actionable message instead of a bare KeyError.

VytCepas:
**Review response (Copilot):** Applied — the test now also asserts `concurrency.group` includes `github.event_name`, so a future refactor can't silently drop it and reintroduce the push/schedule serialization.

VytCepas:
**Review response (Copilot):** Both applied — the test now asserts the exact group string `${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}` (so dropping github.ref, which isolates distinct PRs, also fails), and the doc keeps the cancel-in-progress expression on one line.

PR #774, opened 2026-07-11
feat(PI-761): scaffolded CI resolves matrix per event (floor per-PR, full nightly)

VytCepas:
## Summary

The scaffolded CI ran the **full** Python matrix on every PR. Now it resolves the matrix per event: the **floor** version on a PR/push (fast, cheap feedback), the **full** support window on the nightly `schedule` run — a version-specific break is caught within a day instead of costing N× runner-minutes on every PR. Mirrors what the scaffolder repo does (PI-762), the cost lever requested for the epic.

- **`ci.yml.tmpl`**: one branch in the `resolve` script keyed on `GITHUB_EVENT_NAME` (auto-set by the runner). No new jobs — `lint-and-test` fans out over the resolved set.
- **`test_ci_matrix_per_event`**: extracts and *runs* the resolver under both events, asserting floor-only on `pull_request`/`push` and the full set on `schedule` (mechanism, not prose).
- **`test_templates` resolution tests**: pinned to `GITHUB_EVENT_NAME=schedule` (they test resolution logic, not the trim).
- `quality-gates.md`: CI-cost note records the levers applied — none drops a gate (nightly restores full-matrix + coverage within a day).

Full suite green (2289); actionlint + mkdocs + ruff-format clean.

Closes #761

### DIFF
--- a/docs/development/quality-gates.md
+++ b/docs/development/quality-gates.md
@@ -138,7 +138,15 @@ Documented so their absence is a decision, not an oversight:
 
 ## CI cost note
 
-The repo's required path is a single Python job matrix plus three light jobs (wheel-smoke,
-gitleaks, shellcheck). The advisory scans added under gap 3 run in parallel off the critical
-path and do not extend `ci-gate` wall-clock. Runner-minute deltas from each follow-up PR are
-noted in that PR rather than measured as a separate study.
+The repo's required path is `checks` (lint/typecheck/audit) plus a 4-way sharded `test` job
+(PI-762: pytest-split, race-free, ~5min→~1.5min wall-clock) and three light jobs (wheel-smoke,
+gitleaks, shellcheck). The advisory scans (semgrep, license) run in parallel off the critical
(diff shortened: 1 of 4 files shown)


VytCepas:
**Review response (Codex + Copilot) — all applied:**
- **Restrict full matrix to the nightly cron (Codex P2):** the resolver now keys on `github.event.schedule == '0 3 * * *'` (passed via `GITHUB_EVENT_SCHEDULE`), not `event_name == schedule` — so the weekly Scorecard cron gets floor-only like PR/push. Added `test_full_matrix_only_on_the_nightly_cron` (weekly cron → floor) and `test_resolver_nightly_cron_matches_on_schedule` (drift guard: the hardcoded cron must be a real `on.schedule` entry, else the full matrix would never run).
- **Dedent (Copilot):** already fixed — `textwrap.dedent` on the extracted heredoc.
- **Full-env subprocess (Copilot):** now `{**os.environ, ...overrides}` so uv's proxy/SSL and platform vars survive.
- **Comment wording (Copilot):** updated to say nightly cron specifically.

PR #775, opened 2026-07-11
feat(PI-544): add `project-init doctor` health-check subcommand

VytCepas:
## What

Adds a `project-init doctor [target]` subcommand that health-checks a scaffolded project's `.claude/`/`.agents/` wiring. It runs a fixed checklist, prints `PASS`/`WARN`/`FAIL` with a fix hint per line, and exits non-zero on any `FAIL`. Pure file/JSON inspection — no LLM, no network — matching the scaffolder's deterministic model.

Checks:
- `.agents/config.yaml` exists and carries a parseable scaffold record
- `.claude/settings.json` is valid JSON
- every hook/statusline script it references exists on disk
- referenced `.sh` hooks are executable (`.py` hooks run via the `_py.sh` wrapper, so they don't need `+x`)
- plugin mode: the project-init plugin(s) are enabled; `--no-plugin`: the fallback hooks are present (skip)
- git hooks installed under `.git/hooks/` (WARN before `git init`, not FAIL)
- a Python interpreter is resolvable for the hooks (mirrors `_py.sh`)

## Premise correction

The issue sketched checks against `.claude/hooks/`. Verified against real scaffold output (temp-dir scaffold, both modes): scaffolded hooks live under **`.agents/hooks/`** and are *referenced* from `.claude/settings.json` by `$CLAUDE_PROJECT_DIR` path — there is no `.claude/hooks/` directory in either plugin or `--no-plugin` mode. In plugin mode the guard hooks come from the plugin, so `settings.json` references only the statusline; `--no-plugin` wires all the shell hooks into `settings.json`. The checks follow the real contract, extracting referenced paths from the hook commands (handling the `_py.sh foo.py` two-token form).

## Testing

`tests/integration/test_doctor.py` (11 tests): scaffolds plugin and `--no-plugin` projects, asserts all-PASS + exit 0, then breaks each guarded thing (chmod -x a hook, delete a referenced script, corrupt `settings.json`, delete the config, disable the plugin) and asserts the matching check flips to FAIL and the CLI exits 1. Per the repo rule, each break is paired with a restore to prove the assertion moves. Full suite green (2301 passed); `just code-map` + `just sync-claude` re-run for the new module.

Closes #544

### DIFF
--- a/.agents/docs/CODE_MAP.md
+++ b/.agents/docs/CODE_MAP.md
@@ -48,6 +48,21 @@ Central rich styling for the interactive wizard.
 - `def render_presets` — Print the preset options as an aligned table (plain list off-TTY).
 - `def scaffolding` — Spinner for the duration of a long, single-shot operation.
 
+### `project_init/doctor.py`
+
+``project-init doctor`` — deterministic health-check for a scaffolded project (#544).
+
+- `class Check` — One checklist line: its outcome, a human title, and a fix hint.
+- `def check_scaffold_record` — ``.agents/config.yaml`` exists and carries a parseable scaffold record.
+- `def check_settings_json` — ``.claude/settings.json`` exists and is valid JSON.
+- `def check_referenced_scripts_exist` — Every hook/statusline script settings.json points at is present on disk.
+- `def check_referenced_scripts_executable` — Referenced ``.sh`` hooks carry the executable bit.
+- `def check_plugin_enablement` — Plugin-mode projects enable the project-init plugin(s) in settings.json.
+- `def check_git_hooks` — Git hooks are installed when a ``.github/hooks/`` source ships them.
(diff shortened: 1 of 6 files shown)


VytCepas:
- doctor.py:79 (backfill record vars): Fixing — confirmed real. `_backfill_variables` defaults a pre-ADR-010 record to `no_plugin=true` (those scaffolds were copied-hooks), so I now backfill the parsed variables exactly as `read_scaffold_record` does before the plugin check. Added a test that strips the keys and asserts no false-FAIL.
- doctor.py:285 (git-hook executability): Fixing — confirmed real. Git ignores a non-+x hook, so `check_git_hooks` now WARNs when an installed hook lacks the executable bit instead of reporting PASS. Added a non-executable-hook test.

PR #782, opened 2026-07-11
fix(PI-781): Fresh Rust scaffold is day-one CI red; Go+Rust fail pre-push fast-ci — missing source-existence guards on test/typecheck

VytCepas:
Closes #781

### DIFF
--- a/templates/base/justfile.tmpl
+++ b/templates/base/justfile.tmpl
@@ -206,13 +206,16 @@ ci-local-off:
 # `just --list` shows every recipe. Recipes are thin wrappers — logic lives
 # in the tools and their configs, never in this file.
 
-# download module dependencies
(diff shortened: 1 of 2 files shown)


VytCepas:
Addressed all review feedback in 96cd740:
- **Go `setup` gated on `go.mod`, not `*.go`** (Codex P2): `go mod download` now runs whenever the module is initialized, even with zero `.go` files — verified it syncs the cache in that state.
- **Corrected the go `setup` comment** (setup isn't in the `ci`/`fast-ci` aggregates).
- **Aligned rust `test-cov`/`fuzz` skip messages** to the actionable "run cargo init." wording.
- **Widened `test_day_one_gates.py`** to also cover `format`/`sbom`/`license`/`fuzz` (they short-circuit without a toolchain) and tightened the docstring.

PR #834, opened 2026-07-16
fix(PI-832): rendered agent_guard_adapter.py fails its own ruff format gate

VytCepas:
Closes #832

## What this fixes

The `multi_agent`-gated `agent_guard_adapter.py.tmpl` ends in a standalone
`{{/if multi_agent}}` that carried a trailing newline. When rendered, the
conditional-block substitution left an extra blank line after
`sys.exit(main())`, so a **fresh multi-agent Python scaffold failed `ruff
format --check` (i.e. `just lint`) on day one**. The working sibling
`gen_code_map.py.tmpl` omits that trailing newline — this aligns with it
(one-byte template change).

## Why the existing guard missed it

`test_shipped_python_templates_are_format_clean` runs `ruff format --check`
over raw `templates/`, but ruff skips `.tmpl` files (wrong extension), so a
`.py.tmpl` that renders to malformed Python slips through. Added
`test_rendered_adapter_is_format_clean`, which scaffolds a multi-agent
project and format-checks the **rendered** `.agents/` tree. Verified it
fails on the old template and passes on the fixed one.

## Scope note — the rest of #832

The issue bundled six findings from a fresh `mbdlens` scaffold. On
investigation, only the format-gate one is a project-init bug:

| # | Finding | Verdict |
|---|---------|---------|
| 1 | `dateutil` runtime dep | Not project-init — it emits **no** `pyproject.toml` (recipes short-circuit until the user inits, PI-781). From the reporter's own manifest. |
| 2 | `just` as a pip dep | Same — not in any template/preset. |
| 3 | No `[dependency-groups] dev` | justfile recipes already guard the no-pyproject case; verified day-one green. |
| 4 | Unpinned runtime deps | Not project-init output. |
| 5 | `requires-python = ">=3.14"` | Default `python_floor` is **3.11** (`SUPPORTED_PYTHON_VERSIONS[0]`); the 3.14 was read back from the reporter's existing pyproject. |
| 6 | Scaffolded hook fails `ruff format --check` | **Real bug — fixed here.** |

Verified a fresh `core` scaffold is day-one green on `just lint / typecheck /
test / audit` before this change, and a multi-agent scaffold is green after
it.

https://claude.ai/code/session_01QJvV5TrhFeTTu8NTQtD1p1


VytCepas:
**Review response:**
- [test lacks `cwd=target`, ruff config discovery may use repo config]: Fixing — applied. The check now runs `ruff format --check .` with `cwd=target` so ruff discovers the *scaffolded* project's own `ruff.toml`, mirroring `just lint`. Verified it still passes on the fix and fails on the reintroduced trailing-newline bug.

PR #853, opened 2026-07-17
docs(PI-840): Self-hosted runner guide: document required host binaries (jq missing = exit-127 jobs masquerading as PR failures)

VytCepas:
## Problem

The self-hosted-runner guide documents registration and the `CI_RUNS_ON` switch but not the host binaries the scaffolded workflows assume. Observed live: a fresh WSL runner box was missing `jq`, so `validate-pr` and `board-sync` steps died with exit 127 — reading as a PR-metadata failure, not an environment problem.

## Fix

Audited the workflow templates first: the workflows self-provision most tools via pinned actions (`just`, `shfmt`, language toolchains, gitleaks, semgrep), so the missing piece is exactly the set GitHub's `ubuntu-24.04` images preinstall and a bare host lacks. New **Host prerequisites** section in `self-hosted-ci-runner.md.tmpl`:

- table of assumed binaries — `git`, `curl`, `tar`/`gzip`, `jq`, `shellcheck` (explicitly called out as preinstalled-on-GitHub-images/absent-on-bare-hosts), `gh`, `docker` — each with what uses it and an install hint
- the exit-127 failure signature spelled out so the symptom is searchable
- a copy-pasteable preflight loop (covers the issue's optional `runner_preflight.sh` ask without shipping a new script)
- the `jq`/`gh` lifecycle-workflow rows are `{{#if lifecycle}}`-conditional so they don't dangle in a lifecycle-free scaffold (the `test_lifecycle_none.py` prose scan enforces this)

## Verification

- Two new scaffold-into-tmp tests in `test_local_ci_skill.py`: the lifecycle scaffold's guide names every assumed binary + the preflight; the lifecycle-free guide keeps the section but drops the lifecycle rows. **Proved they can fail**: stashing the template change fails both.
- `test_lifecycle_none.py` prose scan green (33 passed with the module).

Closes #840

https://claude.ai/code/session_017fGBHYbU93F2eidac9GYaR

### DIFF
--- a/templates/base/dot_agents/docs/guides/self-hosted-ci-runner.md.tmpl
+++ b/templates/base/dot_agents/docs/guides/self-hosted-ci-runner.md.tmpl
@@ -65,15 +65,15 @@ Assumed-present binaries:
 | `git` | checkout, every workflow step that touches the repo | `sudo apt install git` |
 | `curl` | download/API steps | `sudo apt install curl` |
 | `tar`, `gzip` | action tool downloads | usually present |
+| `gh` | `ci.yml`'s post-failure PR comment step runs `gh pr comment` on this runner whenever a PR job fails{{#if lifecycle}}; also the lifecycle workflows, if you route them here{{/if lifecycle}} | [cli.github.com](https://cli.github.com) |
 {{#if lifecycle}}| `jq` | JSON parsing in the lifecycle workflows (`validate-pr`, `board-automation`, `issue-validation`, `review-status`) — bites only if you route those here | `sudo apt install jq` |
-| `gh` | lifecycle workflows only — they default to GitHub-hosted, and the PAT-bearing ones must stay there | [cli.github.com](https://cli.github.com) |
 {{/if lifecycle}}| `shellcheck` | the shell-lint step (preinstalled on GitHub images — **absent on bare hosts**) | `sudo apt install shellcheck` |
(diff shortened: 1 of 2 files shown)


VytCepas:
**Review response:**
- *PR modifies monitor_pr.sh, not the runner guide (metadata mismatch)*: Stale — these comments were posted against the pre-rebase head. The branch briefly contained PI-838's commits because `start_issue.sh` branched from the then-checked-out PI-838 branch; it was rebased onto main (force-with-lease) before review, and the current diff is exactly the #840 scope: `self-hosted-ci-runner.md.tmpl` + `test_local_ci_skill.py`. No monitor_pr.sh changes are in this PR.
- *echo lines break mid-phrase ("e.g. a" / "merge proceeds")*: Valid point, but the lines in question are PI-838's and already merged via #852 — not part of this PR. Applying the re-wrap as a separate follow-up fix so this docs PR stays single-scope.

VytCepas:
- *Keep `gh` in the base runner prerequisites (P2)*: Applying — verified the claim at `templates/base/dot_github/workflows/ci.yml.tmpl` (Post failure summary step runs `gh pr comment` under `CI_RUNS_ON`). Moved the `gh` row out of the lifecycle conditional (its description now names the ci.yml failure-comment step; the lifecycle-workflow clause stays conditional) and added `gh` to the preflight loop. The lifecycle-free test now asserts `gh` is present.

PR #860, opened 2026-07-17
fix(PI-847): fix: config.yaml python_floor (3.14) contradicts .python-version / pyproject requires-python (3.13)

VytCepas:
Closes #847


VytCepas:
**Review response:**
- *Unparsable .python-version suppresses the pin render and silently reintroduces drift (P1)*: Applying — verified: `system`/venv-name pins made `_python_floor_from_version_file` return None while the bare `exists()` check still skipped emission, so mise/mypy/CI followed the flag with no visible conflict. The skip is now based on whether the file *supplied* the floor; an unparsable pin renders and engages the `.new`-sibling protection (`.python-version.new` carries the derived floor, the user's file stays untouched). New test: `test_unparsable_pin_surfaces_a_new_sibling_conflict`.

VytCepas:
- *Message claims CI derives from requires-python in both cases*: Fixing — reworded to "the CI matrix follows the declared floor (requires-python when present, else the recorded floor)", accurate whichever file supplied it. (Both Copilot comments on this text addressed together.)
- *upgrade can never propose the pin to older scaffolds*: Applying — the pin is now proposed when the record is a Python project and no .python-version exists, routed through the #249 addition-consent flow as its own `python-pin` group (interactive users get one ask; the scheduled project-init-upgrade workflow already passes `--accept-new all`, so auto-PRs adopt it). An existing unmanaged pin is still never contended for. New tests cover propose-with-consent and never-contend.

VytCepas:
- *agents-dir addition rule unreachable behind claude-core*: Applying — removed the duplicate-prefix rule (pre-existing, surfaced by this diff's context; classifier behavior unchanged since claude-core always won).
- *--accept-new all over-accepts in the migration test*: Applying — accepts only the `python-pin` group now, so any unexpected future addition fails the test.

Issue #935, opened 2026-08-10
just pr-merge can never merge on an org scaffold: --no-review routes to _admin_merge, which refuses under the org profile


VytCepas:
`--no-review` is unusable under the `org` profile: it routes unconditionally to `_admin_merge`, which refuses under exactly that profile. So `just pr-merge` cannot merge anything on an organisation-owned scaffold, and the failure looks like a policy violation rather than a bug.

## What happens

```
$ just pr-merge 96
.agents/scripts/monitor_pr.sh 96 --merge --no-review
PR #96: CI passed. --no-review specified — skipping review gate.
ERROR: admin-merge is refused under the org profile (#251) — hard
  enforcement must bind. Use auto-merge or the merge queue under the
  required checks, or resolve the blocking review/checks.
```

Exit 1. Nothing merges, ever.

## Why it is a bug and not the policy working

The refusal is right in general — admin-merge bypasses server-side rules and must not be a habit. But **skipping a review gate does not require admin**, and the two got conflated.

`monitor_pr.sh:527-530`:

```sh
if [ "$NO_REVIEW" -eq 1 ] && [ "$MODE" = "--merge" ]; then
  echo "PR #$PR_NUMBER: CI passed. --no-review specified — skipping review gate."
  _admin_merge
  exit 0
fi
```

`_admin_merge` (`:327-334`) then refuses whenever `gh_profile` is `org`. There is no other branch, so the two flags cancel each other out and `--no-review` has **no working path at all** on an org-profile repo. The script already carries the correct verb one function up — `_merge_with_retry` (`:305`) does a plain `gh pr merge --squash --delete-branch` with backoff.

The advice in the error message does not apply either. It says to use auto-merge or the merge queue *"under the required checks"* — but the scaffold's own docs record that a **free-plan org cannot have required checks**: branch protection and rulesets both return HTTP 403 (*"Upgrade to GitHub Pro or make this repository public to enable this feature"*). So the suggested remedy is unavailable on precisely the plan where this fires, and there is nothing blocking to "resolve".

## Measured

On the affected repo, two PRs, both with every check green:

| | |
|---|---|
| `gh_profile` | `org` |
| `mergeStateStatus` | `CLEAN` |
| `mergeable` | `MERGEABLE` |
| `reviewDecision` | `""` (empty — CODEOWNERS names the org itself, and GitHub refuses self-approval, so it can never reach `APPROVED`) |
| `GET /branches/main/protection` | `403` |
[...]

PR #940, opened 2026-08-20
fix(PI-932): a fresh scaffold is born green — pyproject.toml, and a secret-scan that can run

### LINKED ISSUE #932: python scaffold: no pyproject.toml template, so `just lint` and `just license` fail on a fresh project
## What happens

A fresh `--language python` scaffold has **red CI on its first push**, before a
line of project code exists. The `lint-and-test` job dies at `just lint`:

```
error: Failed to spawn: `ruff`
  Caused by: No such file or directory (os error 2)
error: recipe `lint` failed on line 18 with exit code 2
```

`license-scan` fails the same way at `just license`. So the first thing a new
Python project does is teach its author that a red badge is normal.

## Why

No `pyproject.toml` template ships:

```
$ find templates -iname "*pyproject*"
$            # nothing
```

The scaffold emits `ruff.toml`, `mypy.ini`, a justfile and a full CI workflow, and
then `uv sync` has nothing to install, so none of the tools those configs govern
exist.

**The asymmetry inside one file is what makes this an oversight rather than a
design choice.** In `templates/base/justfile.tmpl`:

- line 11, `setup` — **explicitly detects the missing file** and degrades:
  `if [ ! -f pyproject.toml ]; then echo "No pyproject.toml yet — nothing to sync."`
- line 18, `lint` — `uv run ruff check .`, unguarded, exit 2
- line 86, `license` — `uv run --with pip-licenses …`, unguarded, exit 1

`setup` was hardened against exactly the state that then breaks the next two
recipes. The TypeScript path is unaffected because its `setup` *installs* the
toolchain (`bun add -d eslint …`), which is why Python is the only language where
a scaffold can be simultaneously complete and unable to run its own gates.

## Proposed fix

**Ship a minimal `pyproject.toml` for `--language python`.** Roughly:

```toml
[project]
name = "{project_slug}"
version = "0.1.0"
description = "{project_description}"
requires-python = ">={python_floor}"
dependencies = []

[dependency-groups]
dev = ["mypy>=1.18", "pytest>=8.4", "pytest-cov>=6.0", "pytest-xdist>=3.6", "ruff>=0.14"]

[build-system]
requires = ["hatchling>=1.25"]
build-backend = "hatchling.build"

[tool.hatch.build.targets.wheel]
packages = ["src/{package_name}"]
```

`{package_name}` is the snake form. That mapping is **already solved** upstream —
`src/project_init/variables.py:547` does `.replace("-", "_")` when rendering
`run_command` — so the same helper covers it and a hyphenated project name needs no
new logic.

Pairing it with `src/{package_name}/__init__.py` and one trivial test would also
give `mypy src/` a target and the coverage gate a denominator, so `just ci` is
green from commit 1 rather than green once somebody has written code.

## The alternative I would argue against

Guarding `lint` and `license` the way `setup` is guarded. It makes CI green, and it
makes a *lint recipe that lints nothing* — a gate reading green while checking
zero files is worse than a red one, because red gets fixed and green gets trusted.
If it is wanted at all, it belongs on the **upgrade** path for existing projects
that have no `pyproject.toml`, never on fresh scaffolds.

## Wizard angle

If the wizard should ask anything here, it is not whether to create the file. Two
questions that would earn their place:

- **src layout or flat** — default `src/`, since the templates already assume
  `src/` in `mypy.ini`, the coverage flag (`--cov=src`) and the CI `CODE_MAP` step.
- **runtime dependencies now or later** — purely so the generated `dependencies = []`
  is a deliberate empty list rather than a thing to discover.

Everything else is derivable from answers the wizard already collects.

## Reproduction

```bash
project-init new demo-py --language python --agents claude --lifecycle none
cd demo-py && just setup && just lint     # setup prints "nothing to sync"; lint exits 2
```

### LINKED ISSUE #933: secret-scan is permanently red on every organisation-owned scaffold (gitleaks-action needs a paid licence)
## What happens

Every scaffold created under a GitHub **organisation** is born with a permanently
red `secret-scan` job. The step fails immediately:

```
GITLEAKS_LICENSE:
##[error]🛑 missing gitleaks license. Go grab one at gitleaks.io and store it as a
GitHub Secret named GITLEAKS_LICENSE.
```

`ci.yml`'s own comment on that step already knows why:

```yaml
# GITLEAKS_LICENSE is required only for organization-owned repos
# (free key from gitleaks.io). Personal accounts need no license.
GITLEAKS_LICENSE: ${{ secrets.GITLEAKS_LICENSE }}
```

So the behaviour is documented — but the consequence is not designed for. A
personal-account scaffold is green and an org-owned scaffold is red, and nothing in
the scaffold or its README tells the org user that their **security** gate needs a
third-party key before it will ever run once.

## Why it matters more than an ordinary red job

This is the one job whose redness is least affordable. A permanently red badge
trains everyone to ignore red — `ci.yml` makes that argument itself, in its
`fail open` comment for the lifecycle checks — and here the thing being ignored is
secret scanning. The failure is also silent in the direction that matters: the job
fails *without scanning*, so a leaked credential in the first commit is not caught
and the red mark looks like the usual licence nag.

## Proposed fix

Run the **binary** instead of the action. It has no licence gate, the ruleset is
identical (`.gitleaks.toml` is already scaffolded), and pinning the version keeps
the gate's strength under version control:

```yaml
      - name: Install gitleaks
        env:
          GITLEAKS_VERSION: 8.30.0
        run: |
          curl -sSfL -o /tmp/gitleaks.tar.gz \
            "https://github.com/gitleaks/gitleaks/releases/download/v${GITLEAKS_VERSION}/gitleaks_${GITLEAKS_VERSION}_linux_x64.tar.gz"
          tar -xzf /tmp/gitleaks.tar.gz -C /tmp gitleaks
          sudo install -m 0755 /tmp/gitleaks /usr/local/bin/gitleaks
          gitleaks version

      - name: Scan full history
        run: gitleaks git --redact --no-banner --verbose .
```

The scaffolded `just scan` recipe already calls the binary
(`gitleaks git --pre-commit --staged --redact --no-banner --verbose`), and the
pre-commit hook does too — so CI is currently the only place in the scaffold that
depends on the licensed action. Switching it makes the three surfaces agree, which
is the stated goal of the agent-agnostic enforcement design.

## If the action is preferred anyway

Then the scaffold should say so where a user will see it before the first push —
one line in the generated README or `CONTRIBUTING.md` under an
`## If your repo is organisation-owned` heading, naming the secret and where the
free key comes from. A comment inside `ci.yml` is not that place; nobody reads a
workflow file before their first push, but everybody sees a red check.

## Reproduction

Create any scaffold under an organisation and push once. `secret-scan` fails on the
first run and on every run after it, with no repository secret set.

VytCepas:
Closes #932
Closes #933

Two fixes to the same class: **a freshly scaffolded project is red before a line of its own code exists.** They ship together because they touch the same two byte-identity fixture files, so splitting them buys a guaranteed rebase conflict and no review benefit.

---

## PI-932 — a python scaffold ships no `pyproject.toml`

Reproduced first, on a real scaffold rather than from the ticket:

```
just setup   No pyproject.toml yet — nothing to sync.
just lint    error: Failed to spawn: `ruff`      (rc=2)
```

The scaffold emits `ruff.toml`, `mypy.ini`, a justfile and a full CI workflow, then `uv sync` has nothing to install, so none of the tools those configs govern exist. `templates/base/pyproject.toml.tmpl` is `{{#if python}}`-gated the way `justfile.tmpl` is; a node scaffold gets no `pyproject.toml` (asserted).

The alternative — guarding `lint` and `license` the way `setup` is guarded — is argued down in the template header: a lint recipe that lints nothing reads green while checking zero files, and green gets trusted where red gets fixed.

**Two corrections to the ticket, both measured.** `just license` does *not* fail the same way — `uv run --with pip-licenses` installs the tool on the fly, rc=0 before this change; only `lint` was red. And the reproduction command is stale: there is no `new` subcommand, and `--non-interactive` requires `--preset`.

**A second defect this uncovered**, and the fix is incomplete without it: once ruff existed, `ruff format --check` went red on the scaffolded `.agents/rules/python.md`, whose Hypothesis example is not ruff-formatted. Invisible while ruff could not be spawned at all.

`package_name` is **one function**, not a fourth inline `.replace("-", "_")`. The conversion already lived inside `render_run_command`; the template needs it for the package directory and so do both upgrade/migration builders. Four copies that must agree is the drift this repo keeps finding.

`tests/helpers.py:make_variables` is a hand-maintained second copy of the variable contract — adding one key turned **331 tests red** until the helper learned about it. Fixed here; the copy itself deserves its own ticket, and its `run_command` default is already drifted from what the real builder produces.

## PI-933 — `secret-scan` is permanently red on every org scaffold

`gitleaks/gitleaks-action` refuses to run on an organisation-owned repo without a `GITLEAKS_LICENSE`. It is the one job whose redness is least affordable: it fails **without scanning**, so a credential in the first commit is not caught while the red mark reads as the usual licence nag.

The binary has no licence gate, reads the same `.gitleaks.toml`, and is already what `just scan` and the pre-commit hook run — CI was the only surface on the licensed action. It also **retires the TMPDIR bug** the old step carried a paragraph about.

**Three things the ticket's snippet did not have**, each a defect if omitted:

| | why |
|---|---|
| OS/arch read from `RUNNER_OS`/`RUNNER_ARCH` | `runs-on` is overridable and the runner that found the TMPDIR bug was self-hosted **macOS** — a hardcoded `linux_x64` asset 404s there and the security gate fails for a reason unrelated to secrets |
| checksum verification before execution | replacing a SHA-pinned action with a curl is a supply-chain downgrade unless the bytes are checked. Honest about the limit in the comment: it catches corruption, truncation and CDN substitution, not a compromised release — that is what the pin is for |
| a `renovate:` marker **and the customManager that reads it** | no built-in manager sees a version literal in a shell step, so a pinned scanner would freeze while everything else stayed current. A security tool rotting quietly is worse than an unpinned one. The first draft's comment claimed Renovate maintained the pin; the manager now makes that true rather than deleting the claim |

Also 8.30.1, not the ticket's 8.30.0, which is not the current release.

---

## Evidence — executed, not read

```
fresh python scaffold, every gate      lint 0 · typecheck 0 · test-cov 0 · license 0 · ci 0
node scaffold                          no pyproject.toml            (negative control)
install step, macOS/ARM64              checksum OK, gitleaks 8.30.1 (run for real)
tampered tarball                       verify FAILED, rc=1
scan, clean scaffold                   rc=0, "no leaks found"
scan, planted PAT + AWS key + RSA key  rc=1, 3 findings, all REDACTED
renovate regex vs the real marker      github-releases / gitleaks/gitleaks / 8.30.1
suite                                  2765 passed, 6 failed
baseline on ff8f28e                    2761 passed, 6 failed — the SAME six
```

**Zero regressions.** `main` is already red by six tests unrelated to this change (`test_observability_self_log`, `test_post_edit_lint_autofix` ×3, `test_post_edit_lint_typecheck` ×2).

**Every new assertion was watched go red.** Five mutations of the CI template: no checksum download, no renovate marker, hardcoded linux/x64, verify moved after the extract, licensed action put back. **Two assertions were rewritten because their first cut stayed green under mutation** — `checksums.txt` survived in the `grep` that consumes it, and the runner-OS variable survived in the unsupported-value error message. Both now assert the fetch and the dispatch instead of the name.

One probe of mine was also wrong and is worth recording: the first planted secret was the canonical AWS **documentation** key, which gitleaks correctly allowlists, so the scan reported clean. That was a bad fixture, not a blind gate — realistic secrets are caught.

---

## Two findings for the operator, not fixed here

- **This repo's own CI workflow still uses the licensed action.** It works today because `VytCepas` is a personal account; it becomes the identical permanent red the day this repo moves under an organisation. It is a required check here, so changing it is a separate decision. Recorded in `docs/development/quality-gates.md`.
- **#935 is already fixed** and can be closed. PI-936 (`ff8f28e`) rerouted `--no-review` through `_merge_with_retry` for a `CLEAN`/`UNSTABLE` PR and left `_admin_merge` for everything else — exactly #935's suggested fix.


VytCepas:
**Review response:** both findings accepted and fixed in 779479c.

- **`grep " ${asset}$"` is a regex, so the dots are wildcards** — Fixing. The line reads as an exact selection and is not one. I checked whether it can actually mis-select today: no sibling asset in gitleaks' manifest differs from another only at a dot, so it cannot — which is the reason to fix it rather than not, because that makes it a defect that surfaces only when a future asset name changes, long after anyone remembers this line. Now `awk -v a="$asset" '$2 == a { print; found = 1 } END { exit !found }'`, which compares the field for equality and prints the line verbatim so `shasum -c` still gets the format it expects. `exit !found` keeps the other half honest: selecting nothing must abort, because an empty `expected.sha256` handed to a verifier is a check that reports success on no input.

  Verified by running the step, not by reading it: the real download verifies OK and selects exactly one line; an asset name absent from the manifest exits 1; a wildcard-shaped name exits 1 against `awk` while the old `grep` pattern matched. Reverting `awk` back to `grep` in the template turns the new assertion red.

- **`_scan_job()` raises IndexError when a delimiter moves** — Fixing. Correct, and the ordering point is the sharper half: pytest does not guarantee `test_secret_scan_job_present` runs first, so the whole class failed with a traceback about list indices rather than "the job is gone". Both delimiters are now asserted with a message. Renaming the job in the template now reports `no secret-scan job in the scaffolded CI` instead of an IndexError.

Suite after the fixes: 2765 passed, 6 failed — the same six that already fail on `main` (`test_observability_self_log`, `test_post_edit_lint_autofix` ×3, `test_post_edit_lint_typecheck` ×2), none related to this PR. Zero regressions, `just lint` rc=0.

VytCepas:
Both review threads resolved — each was addressed in 779479c, not waved through. Nudging `review/decision` to recompute (resolving a thread emits no workflow event; this comment is the documented `issue_comment` trigger).

PR #944, opened 2026-08-21
fix(PI-939): a bump PR that can report, and a review status that unsticks itself

### LINKED ISSUE #939: Scheduled bump PRs open with github.token, so every one of their checks lands at action_required and the PR can never go green on its own
`third-party-updates.yml` opens its bump PRs with `GH_TOKEN: ${{ github.token }}` (`.github/workflows/third-party-updates.yml:34`). Every workflow those PRs trigger is queued as **`action_required`** and waits for a human click that nothing asks for, so the PR sits at `mergeStateStatus: BLOCKED` with **zero check runs on its head SHA** — not red, not green, simply never reported. The update automation therefore produces PRs that can never merge themselves, and they accumulate silently.

## Measured

Two PRs, both from the scheduled workflow:

| | #934 (ccr 3.0.20) | #938 (ccr 3.0.21) |
|---|---|---|
| created | 2026-08-10 | 2026-08-17 |
| author | `app/github-actions` | `app/github-actions` |
| workflow runs | 4 × `completed/action_required` | 4 × `completed/action_required` |
| check runs on head SHA | **none** | **none** |
| `mergeStateStatus` | `BLOCKED` | `BLOCKED` |
| required contexts reporting | 0 of 3 | 0 of 3 |
| days idle before intervention | 10 (closed unmerged, superseded) | 3 |

The three required contexts are `Check PR title format`, `Check for Closes keyword in PR body` and `CI gate`. None of them can report while the run that produces them is unapproved, so branch protection is unsatisfiable by construction rather than by any property of the change.

**This is a regression, not the standing behaviour.** #905 (2026-08-02) came from the same bot author on the same workflow and collected all 20 checks green — `CI gate`, `Tests (shard 1-4/4)`, `Semgrep`, `Secret scan (gitleaks)`, `Dependency license scan`, both portability jobs, `Wheel smoke test` — then merged normally. Something changed in the approval policy between 2026-08-02 and 2026-08-10; the repo-level Actions approval setting is not exposed over REST (`/actions/permissions/access` returns 422 for public repos), so this needs a look in Settings → Actions → General.

## The change was never the problem

Once the four runs on #938 were approved by hand (`POST /repos/{owner}/{repo}/actions/runs/{id}/approve`), all 20 checks ran and passed inside two minutes, and a requested Copilot review returned *"reviewed 2 of 2 files, generated no comments."* The PR was a two-line version-string bump in lockstep across `tools/pinned_third_party.toml` and `templates/multi_model/dot_agents/scripts/setup_models.sh`.

## Second-order trap: one run cannot be approved at all

After the review landed, the `Review status` run triggered by the `pull_request_review` event was **also** `action_required`, and the approve endpoint refuses it:

```
POST /repos/{owner}/{repo}/actions/runs/32364610938/approve
403  This run is not from a fork pull request or queued by the Actions bot
```

So `review/decision` stays `pending` with no API route to clear it. What worked is the fallback the workflow's own header documents — the `issue_comment` trigger: a plain comment on the PR re-ran the status job and it settled to `success — "Reviewed — no open comments"`. Worth knowing, because the obvious remedy (approve the run) is unavailable on exactly the event that gates the merge, and the non-obvious one is a comment.

## Suggested fix

1. Open the PR with a token that is not `GITHUB_TOKEN` — a PAT or a GitHub App installation token on the create step. PRs authored that way trigger workflows normally, which is the documented reason `GITHUB_TOKEN`-created events are throttled in the first place.
2. Failing that, have the workflow approve its own pending runs after creating the PR, or narrow the repo's Actions approval policy so same-repo bot PRs are exempt.
3. Either way the acceptance test is the same and is cheap to run: let one scheduled bump PR open, and assert its head SHA carries check runs without a human touching anything.

Until then every bump PR needs a manual "Approve and run", and a comment afterwards to settle `review/decision`.

## Related

#870 (do not bump the ccr pin mechanically) is about the 2.x → 3.x migration and is unaffected by this — it is the *review* of the version, not the mechanics of the PR. #933 and #935 are the same family of defect one layer out: a control that works on one profile and is inert or blocking on another.

VytCepas:
Closes #939

## The root cause is narrower than the ticket, and it changes the fix

Re-measured rather than taken from the ticket. Every run currently sitting at `action_required` in this repo:

```
2026-08-21  Review status  actor=Copilot   branch=feat/PI-893-...
2026-08-21  Review status  actor=Copilot   branch=feat/PI-893-...
2026-08-20  Review status  actor=Copilot   branch=fix/PI-932-...
2026-08-11  Review status  actor=Copilot   branch=fix/PI-936-...
```

Not `github-actions[bot]` — **`Copilot`**. The approval policy fires on *bot actors generally*, so it hits two things: the bump PRs the ticket is about, **and** the `pull_request_review` run that computes `review/decision` on every ordinary PR. That second one is why `review/decision` had to be nudged by hand on all three PRs merged today. The ticket recorded it as a "second-order trap"; it is the same defect, and it is the half that costs something on every PR rather than weekly.

Two of the ticket's claims re-derived, both still true:

```
GET  /actions/permissions/access                422  "only applies to internal and private repositories"
POST /actions/runs/32452514858/approve          403  "not from a fork pull request or queued by the Actions bot"
```

## What ships

**The bump PR** — opened under `secrets.BUMP_PR_TOKEN` when it exists, `github.token` when it does not. The secret is optional on purpose: a missing one must not turn the weekly check into a red build. The fallback emits a `::warning::` naming `action_required`, so the resulting stuck PR carries its own explanation instead of accumulating silently.

**A best-effort recovery** — `tools/approve_pending_bump_runs.sh` approves what it can and **says what it cannot**. The 403 above means the `pull_request_review` run is not recoverable this way, and a remedy that silently covers half the problem is worse than one that names its half. It exits 0 unconditionally: turning "could not un-stick a PR" into a failed weekly check replaces a quiet problem with a noisy one. It only ever touches PRs authored by `app/github-actions`, because approving a run is running code.

**`monitor_pr.sh` nudges `review/decision` itself.** This is the part with daily value. The status is computed by a workflow, and both events that should refresh it fail to — resolving a thread emits no Actions event at all (`pull_request_review_thread` is webhook-only, #719), and the bot's review run queues at `action_required` with no API route to clear it. `issue_comment` *is* a trigger on that workflow and *is* attributed to the commenter, so a plain comment settles it. That was a manual step on every PR; it is not any more.

## What this does NOT fix, plainly

**The approval policy itself.** It is a repository setting (Settings → Actions → General), it is not readable or writable over REST for a public repo (the 422 above), and narrowing it is the only change that removes the problem rather than working around it. Everything here is a workaround, and the docs say so.

**The acceptance test cannot run in this PR.** #939's own criterion is "let one scheduled bump PR open and assert its head SHA carries check runs without a human touching anything". That needs a real upstream release and the operator to add `BUMP_PR_TOKEN` first. So the workflow half is verified as *wired*, not as *working* — the tests for it are text assertions on a YAML file, which is the weakest kind of test, and their docstring says exactly that.

**Two operator actions are required** and neither is in my reach: add the `BUMP_PR_TOKEN` secret, and consider narrowing the approval policy. Both documented in `docs/development/repo-reference.md`.

## Evidence

```
new nudge tests        2 passed
workflow contract      5 passed
full suite             2949 passed, 6 failed — the same six that fail on main
lint                   ruff check + format --check, rc=0; shellcheck --severity=warning clean
approve script         runs clean against this repo (no bot PRs → exit 0)
run discovery          verified against a real head SHA carrying an action_required run
```

The nudge was broken and watched go red — **4 mutants, 0 survivors**: the call removed, the comment call neutered, the already-green short-circuit forced off, the status read blanked. Both the template and the synced `.agents/` copy restored byte-identical afterwards, asserted.

The nudge is deliberately **not** unconditional: a PR whose `review/decision` is already green gets no comment, and that negative case is pinned. Commenting on every PR would be noise, and noise trains a reader to skim past the comment that mattered.

## One thing found on the way, worth a follow-up

`test_invokes_gitleaks_and_propagates_findings_exit` fails whenever **this repo's own lint is red** — the scaffolded pre-commit hook runs `just lint` as gate 1 and exits before ever invoking gitleaks, so the assertion that fails is `"--staged" in log.read_text()` with a `FileNotFoundError`. It reports a gitleaks problem that is not one. It cost me a wrong diagnosis earlier today, twice. Not fixed here — it is a test-diagnostic defect, not a gate defect, and it belongs in its own change.

### DIFF
--- a/.agents/scripts/monitor_pr.sh
+++ b/.agents/scripts/monitor_pr.sh
@@ -210,6 +210,64 @@ sys.exit(len(bad))
 " "$IGNORE_CHECKS"
 }
 
+# PI-939: read the derived review/decision commit status, or "" if absent.
+_review_decision_state() {
+  gh pr checks "$PR_NUMBER" --json name,bucket 2>/dev/null | "$PY" -c "
+import json, sys
+try:
+    data = json.load(sys.stdin)
+except Exception:
+    sys.exit(0)
+for c in data:
+    if c.get('name') == 'review/decision':
+        print(c.get('bucket') or '')
+        break
+" 2>/dev/null || true
+}
+
+# PI-939: re-trigger review-status.yml with a plain PR comment.
+#
+# review/decision is computed by a workflow, and the two events that SHOULD
+# refresh it both fail to. Resolving a review thread emits no Actions event at
+# all (`pull_request_review_thread` is a webhook event only, #719). And where
+# the repository's Actions policy requires approval for bot actors, the
+# `pull_request_review` run a bot reviewer triggers is queued at
+# `action_required` — which cannot be cleared from the API either:
+#
+#   POST /actions/runs/{id}/approve
+#   403  This run is not from a fork pull request or queued by the Actions bot
+#
+# So the status stays stale on a PR whose review gate has actually passed, and
+# the merge below reports BLOCKED with no visible cause. `issue_comment` IS a
+# trigger on that workflow and IS attributed to the commenter rather than a
+# bot, so a comment settles it. That was a manual step; it is not any more.
+_nudge_review_decision() {
+  local state elapsed=0
+  state=$(_review_decision_state)
+  if [ "$state" = "pass" ]; then
+    return 0
+  fi
+  echo "review/decision is '${state:-absent}' though the review gate passed — commenting to re-trigger review-status.yml (PI-939)."
+  if ! gh pr comment "$PR_NUMBER" \
+    --body "Review threads addressed and resolved. Re-triggering \`review/decision\`: resolving a thread emits no workflow event, so this comment is the documented nudge." \
+    >/dev/null 2>&1; then
+    echo "  could not post the comment — review/decision left as-is."
+    return 0
+  fi
+  while [ "$elapsed" -lt 120 ]; do
+    sleep 15
+    elapsed=$((elapsed + 15))
+    state=$(_review_decision_state)
+    if [ "$state" = "pass" ]; then
+      echo "  review/decision: pass after ${elapsed}s."
+      return 0
+    fi
+  done
+  # Not an error: the status may legitimately be failing, and the merge step
+  # below reports that with its own diagnostics. Say what was tried.
+  echo "  review/decision still '${state:-absent}' after ${elapsed}s — the merge may report BLOCKED."
+}
+
 # Print review feedback — inline comments first, falls back to full PR comments view.
 _print_review_comments() {
   local inline
@@ -721,6 +779,14 @@ fi
 
 PR_URL=$(gh pr view "$PR_NUMBER" --json url -q '.url')
 echo "PR #$PR_NUMBER passed: $PR_URL"
+
+# The review gate above has passed. If the derived status has not caught up,
+# nudge it before attempting the merge — not after, when the failure mode is an
+# opaque BLOCKED. Skipped when the operator turned the review gate off, because
+# there is then no gate whose result the status is lagging behind.
+if [ "$MODE" = "--merge" ] && [ "$MAX_REVIEW_CYCLES" -ne 0 ] && [ "$REVIEW_DECISION" != "SKIPPED" ]; then
+  _nudge_review_decision
+fi
 
 if [ "$MODE" = "--merge" ]; then
   MERGE_STATE=$(gh pr view "$PR_NUMBER" --json mergeStateStatus -q '.mergeStateStatus' 2>/dev/null || echo "UNKNOWN")
--- a/.claude/scripts/monitor_pr.sh
+++ b/.claude/scripts/monitor_pr.sh
@@ -210,6 +210,64 @@ sys.exit(len(bad))
 " "$IGNORE_CHECKS"
 }
 
+# PI-939: read the derived review/decision commit status, or "" if absent.
+_review_decision_state() {
+  gh pr checks "$PR_NUMBER" --json name,bucket 2>/dev/null | "$PY" -c "
+import json, sys
+try:
+    data = json.load(sys.stdin)
+except Exception:
+    sys.exit(0)
+for c in data:
+    if c.get('name') == 'review/decision':
+        print(c.get('bucket') or '')
+        break
+" 2>/dev/null || true
+}
+
+# PI-939: re-trigger review-status.yml with a plain PR comment.
+#
+# review/decision is computed by a workflow, and the two events that SHOULD
+# refresh it both fail to. Resolving a review thread emits no Actions event at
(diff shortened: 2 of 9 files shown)


VytCepas:
**Review response:** both findings accepted and fixed. The P1 is the more serious of the two and it defeated a boundary this script's own header claims to respect.

- **P1 — an explicitly named PR bypassed the author filter.** Correct. The filter was applied during *discovery* only, so `approve_pending_bump_runs.sh 123` reached the approve call for any PR at all — and approving a run **runs the code in that run**, which is exactly the thing the header says must never happen on a PR a human or a third party opened. A guard that holds on the default path and not the explicit one is not a boundary; it is a habit.

  The check now runs inside the loop, per PR, whatever the source. Verified against real human-authored PRs rather than a stub:

  ```
  $ bash tools/approve_pending_bump_runs.sh 944 942
  PR #944: author is 'VytCepas', not app/github-actions — refusing to approve its runs.
  PR #942: author is 'VytCepas', not app/github-actions — refusing to approve its runs.
  approve_pending_bump_runs: 0 approved, 0 refused.
  ```

- **P2 — one query, and run creation is asynchronous.** Also correct, and the failure mode is the nastiest kind: the script would print "no runs waiting" about a PR that was *about to be* stuck, which reads as success. It now polls (90s default, 10s interval, both env-overridable so the tests do not sleep).

  The poll needed a stop condition that is not just a timeout, or the happy path would cost 90 seconds on every run. It exits early on either answer: runs to approve, **or** runs reporting normally. And the two no-op outcomes are no longer worded the same — `N run(s) reporting normally` versus `NO workflow runs appeared within Ns — that is the stuck state, not a healthy one`. Wording is load-bearing here: "nothing to do" is what let two bump PRs accumulate unnoticed in the first place.

**Both fixes were broken and watched go red** — 4 mutants, 0 survivors: the in-loop author check disabled, the polling loop collapsed to a single pass, the stuck/healthy branch forced to the healthy wording, and the approve call neutered. Script restored byte-identical afterwards, asserted.

One thing I got wrong on my own and fixed: the first cut of the *stub* shared a round counter between the run-discovery and total-count branches, so it answered inconsistently about the same moment and the polling test failed for a reason that had nothing to do with the script. A stub that lies is worse than no test; it reads as evidence.

```
new approver tests   5 passed
workflow contract    5 passed  (one assertion updated — it pinned a literal the fix replaced)
full suite           2954 passed, 6 failed — the same six that fail on main
lint                 rc=0; shellcheck --severity=warning clean
```

VytCepas:
Review threads addressed and resolved. Re-triggering `review/decision`: resolving a thread emits no workflow event, so this comment is the documented nudge.

Issue #870, opened 2026-07-20
migrate multi-model overlay to CCR 3.x (sqlite config, Node 22 floor) — do not bump the pin mechanically


VytCepas:
## Summary

migrate multi-model overlay to CCR 3.x (sqlite config, Node 22 floor) — do not bump the pin mechanically

## Metadata

- Type: chore
- Priority: high
- Size: L
- Agent ready: No
- Confidence: medium
- Scale: task
- Area: templates/multi_model

## References

- https://github.com/musistudio/claude-code-router/releases

## Dependencies

- #869

## Acceptance criteria

- [ ] Define acceptance criteria before implementation

## Definition of Ready

- [ ] Acceptance criteria are clear enough to verify
- [ ] Dependencies, references, and affected areas are recorded

## Definition of Done

- [ ] Implementation is complete
- [ ] Relevant checks, tests, or manual validation are documented

## Context

`tools/pinned_third_party.toml` pins `ccr` (`@musistudio/claude-code-router`) at **2.0.0**. The scheduled `third-party-updates` workflow reports **3.0.7** available.

**Do not apply this bump mechanically.** `check_third_party_updates.py apply` only rewrites the `CCR_VERSION` string; CCR 3.x is a breaking change that requires reworking `setup_models.sh`.

## Why 3.x is not a drop-in

1. **Config moved from JSON to SQLite.** v3 reads `~/.claude-code-router/config.sqlite`. A legacy `config.json` is read **once for migration only**, then ignored. This invalidates the whole `seed_config()` function in `setup_models.sh` — re-running to update keys becomes a silent no-op after first run, and the merge-safe backup/overwrite prompt no longer describes reality.
2. **Node 22+ required.** New runtime floor; the script does not check it.
3. **No release notes for 3.0.0-3.0.2.** A major bump on the request path with no documented migration path.
4. **npm lags GitHub by 8 releases.** npm `dist-tags.latest` = 3.0.7 (2026-07-03); GitHub is tagged through 3.0.15 (2026-07-20). Bumping to 3.0.7 lands on a version whose known gateway/routing bugs are fixed only in releases not published to npm. Verify npm has caught up before choosing a target version.
5. **Architecture shift.** v3 introduces a gateway layer (`@the-next-ai/ai-gateway`), a desktop app, and local-agent OAuth import — the `ccr ui` / `ccr code` surface the scaffolded guide documents may have moved.

[...]

PR #966, opened 2026-08-31
chore: bump ccr 3.0.21 → 3.0.22

VytCepas:
Automated proposal to bump the vetted pin for **ccr** (`@musistudio/claude-code-router`).

- **3.0.21 → 3.0.22**
- Security scan — npm audit: 0 vulnerabilities (critical 0, high 0, moderate 0, low 0).
- Changelog: https://github.com/musistudio/claude-code-router/releases

This sits on a scaffolded project's request path, so **review the changelog/diff
before merging** (ADR-016 §5). Not auto-merged. Downstream projects inherit the
vetted pin via upgrade-as-PR (PI-241).

### DIFF
--- a/templates/multi_model/dot_agents/scripts/setup_models.sh
+++ b/templates/multi_model/dot_agents/scripts/setup_models.sh
@@ -21,7 +21,7 @@ set -euo pipefail
 # hold every scaffolded project behind a bump PR on a fast-moving tool. It is
 # installed at upstream latest and is absent from the manifest by design.
 CCR_PKG="@musistudio/claude-code-router"
-CCR_VERSION="3.0.21"
+CCR_VERSION="3.0.22"
 CLAUDE_PKG="@anthropic-ai/claude-code"
 
 # --- paths --------------------------------------------------------------------
--- a/tools/pinned_third_party.toml
+++ b/tools/pinned_third_party.toml
@@ -10,7 +10,7 @@
 [tools.ccr]
 package = "@musistudio/claude-code-router"
 ecosystem = "npm"
-pinned = "3.0.21"
+pinned = "3.0.22"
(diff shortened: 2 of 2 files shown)


VytCepas:
## Recommend holding this until #870 lands

Not a review of the diff — the bump itself is mechanically correct. The objection is that the target it bumps has not been migrated.

**#870** exists specifically to say *"do not bump the pin mechanically"*, because CCR 3.x moves its config store from JSON to SQLite. That pin has since crossed the major boundary twice anyway (#905 `2.0.0 → 3.0.16`, #938 `3.0.18 → 3.0.21`) and the overlay was never reworked:

```
$ grep -rn "sqlite" templates/multi_model/
(no output)

setup_models.sh:40   GLOBAL_CONFIG="$GLOBAL_DIR/config.json"
models.sh:14         CONFIG="${CCR_CONFIG:-$HOME/.claude-code-router/config.json}"
setup_models.sh:68   die "need bun or npm (Node >=20)"     # #870 §2 says the v3 floor is 22
```

So this PR is the third mechanical bump on a seeder that, per #870's research, writes a file CCR 3.x reads only once and then ignores.

**This PR does not make anything worse** — `3.0.21 → 3.0.22` is a patch bump within a major the overlay already fails to support. The reason to hold is narrower: merging it a third time keeps recording "the pin is maintained" while the surface underneath it stays unmigrated, and that is the signal that let 3.0.16 and 3.0.21 through unremarked.

The `review/decision` check is green (`Reviewed — no open comments`) and the branch is **behind** `main`, so it needs a base merge before it could go in regardless.

Not closing it — that is the maintainer's call, and the bump is still wanted *after* the migration. Flagging so it is a decision rather than a default.

Full evidence in https://github.com/VytCepas/project-init/issues/870#issuecomment-5559729622.

---

PR #972, opened 2026-09-06
fix: the payload hash counted gitignored bytecode, so a local hook run reddened CI's green

VytCepas:
`payload_sha256` walks `plugin_root.rglob("*")` and hashes every file it finds. `plugins/project-init-workflow/hooks/prod_guard.py` is an importable module, so importing it writes `__pycache__/prod_guard.cpython-*.pyc` **inside the payload dir**. Those are gitignored, so a clean CI checkout never has them and the committed lock is computed without them.

## What was measured

On this box, 2026-09-06: two stray `.pyc` (cpython-311 and cpython-314) made `test_lock_matches_the_committed_tree` fail locally while the same test at the same SHA — 6bfe80b, whose CI run is green — passed on the runner. `rm -rf` on that one directory turned the suite green with nothing else changed.

The test's own name is the argument: it asserts the lock matches the **committed** tree, and untracked bytecode is not in it. A tripwire that fires on an artefact of running the very thing it guards is a false positive — the operator learns to ignore a red that will one day be real payload drift, which is the drift PI-881 exists to catch.

## Third place the walk had to learn this

`scaffold.py:49` and `scaffold.py:861` already carry the exclusion, the second with `.pyo` as well. This matches both, so a scaffolded project and the plugin payload now agree on what "the product" means.

## No version bump is owed

The committed lock was computed on CI without the bytecode, so this makes the local value match what is already recorded. `sync_plugin.py --check` reports `plugin payload is in sync`; `plugins/payload-locks.json` is untouched.

## Mutation-tested

With the two `and` clauses reverted, `test_bytecode_next_to_a_hook_does_not_move_the_hash` fails **by name**; restored, it passes. The test also asserts the exclusion is *not* a blanket one — adding a real `h2.py` in the same directory must still move the digest — because a walk that returned nothing would satisfy the first assertion just as well.

Verified: 3098 passed, 6 skipped; `ruff check` and `ruff format` clean.

### DIFF
--- a/tests/contracts/test_plugin_marketplace.py
+++ b/tests/contracts/test_plugin_marketplace.py
@@ -210,6 +210,35 @@ class TestPayloadVersionLock:
             for hook in (root / "hooks").glob("*.sh"):
                 assert hook.stat().st_mode & 0o111, f"{root.name}/{hook.name} is not executable"
 
+    def test_bytecode_next_to_a_hook_does_not_move_the_hash(self, tmp_path):
+        # THE HASH IS OVER THE COMMITTED TREE, AND BYTECODE IS NOT IN IT.
+        # `hooks/prod_guard.py` is importable, so any local run of the guard
+        # writes `__pycache__/*.pyc` inside the payload dir. Measured 2026-09-06:
+        # two such files made test_lock_matches_the_committed_tree fail locally
+        # while the same SHA was green on CI, whose checkout has none. A red that
+        # ordinary local work manufactures is No.2.11's false positive, and the
+        # operator learns to ignore the one that will later be real drift.
+        (tmp_path / ".claude-plugin").mkdir()
+        (tmp_path / ".claude-plugin" / "plugin.json").write_text('{"version": "1.0.0"}')
+        hook = tmp_path / "hooks" / "h.py"
+        hook.parent.mkdir()
+        hook.write_text("x = 1\n")
+        clean = sync_plugin.payload_sha256(tmp_path)
+
+        # Both spellings the walk has to survive: a __pycache__ dir, and a stray
+        # .pyc sitting directly beside the source.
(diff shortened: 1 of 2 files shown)


VytCepas:
Both review findings taken — the Codex P2 was a real defect and in the unsafe direction.

**Codex P2 (fixed in babfc8e).** `"__pycache__" not in p.parts` asks about the whole absolute path, so a `plugin_root` beneath an ancestor of that name matches on *every* candidate: `_payload_files` returns `[]`, `payload_sha256` becomes the digest of nothing, the lock agrees with itself forever, and PI-881 stops detecting the stale-plugin drift it exists for. Silent, and failing open. The `.claude-plugin` clause on the line directly above already used `p.relative_to(plugin_root).parts` — the new clause just didn't match its neighbour. Both are plugin-relative now.

Added `test_an_ancestor_named_pycache_does_not_empty_the_payload`, which asserts both halves: the walk still finds the hook under such an ancestor, **and** the digest equals what the same tree produces elsewhere, so a checkout's location cannot move the lock. Mutation-tested — restoring the absolute-path form fails it by name.

**Copilot nit (fixed in the same commit).** The docstring hardcoded `scaffold.py:49` / `scaffold.py:861`; those drift while still reading correctly. It now names `_tree_digest` and `_iter_layer_sources` instead, which is the same reason CLAUDE.md gives for deriving a fact rather than copying it.

Verified after both: 3099 passed, 6 skipped; `ruff check` and `ruff format` clean; `sync_plugin.py --check` still reports `plugin payload is in sync` with `payload-locks.json` untouched.

VytCepas:
Both review threads resolved — the Codex P2 fix and the Copilot docstring nit are in babfc8e, with a mutation-tested assertion for the first. Commenting to nudge `review/decision`, which this repo's own workflow notes only refreshes on a following event.

